PR #76, opened 2020-09-25
Complete tasks supplied by PR author on merge

owenauch:
Per suggestion in https://app.asana.com/0/406675932871260/1191254163824112/f, this PR causes SGTM to look for task ids after the string "Tasks to complete on merge:" in the PR body, and complete them when the PR merges. I still haven't figured out where logic to add these task ids / urls to the description should live. I think git pr should prompt for tasks to complete on merge but I'm not sure yet if it makes more sense to insert to the description in git pr or SGTM.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1195893131721939)


src/asana/helpers.py
```diff
@@ -309,6 +309,32 @@ def urlreplace(matchobj: Match[str]) -> str:
...
+    body_lines = pull_request.body().splitlines()
+    stripped_body_lines = (line.strip() for line in body_lines)
+    task_url_line = None
+    seen_asana_tasks_line = False
+    for line in stripped_body_lines:
+        if seen_asana_tasks_line:
+            task_url_line = line
+            break
+        if line.startswith("Asana tasks:"):
+            seen_asana_tasks_line = True
```
owenauch, line 325: If there's a cleaner way to do this let me know — I experimented for a bit and this was the best I could do

owenauch (commented):
@gregeinfrank This has changed a decent bit. The main differences:
* I'm going to use a label to control this functionality, similar to automerge, which should simplify the patterns here.
* @kasrakoushan is changing `git pr` to explicitly prompt for multiple task urls, which will be in a section of the description under a line that says "Asana tasks:", so we can use that to find the task urls.
* Once Kasra's PR lands (he had to revert on Friday due to a bug), I'll add a prompt in `git pr` to ask if you want to complete linked tasks on merge, which will set the label. cc @binoche9 

I QAed the auto-complete functionality with https://github.com/Asana/codez/pull/101687 and it worked properly.

src/asana/logic.py
```diff
@@ -0,0 +1,23 @@
...
+    COMPLETE_ON_MERGE = "complete tasks on merge"
+
+
+def should_autocomplete_tasks_on_merge(pull_request: PullRequest):
+    return (
+        _is_autocomplete_feature_enabled()
+        and pull_request.merged()
+        and pull_request_has_label(
+            pull_request, AutocompleteLabel.COMPLETE_ON_MERGE.value
+        )
```
owenauch, line 18: Is there any need to include a check in here to make sure the task isn't already completed? I'd imagine it wouldn't matter if we update it to be completed and it already is.

binoche9: Maybe check whether the API creates another story/notification if you try to complete a task that is already completed?

src/asana/helpers.py
binoche9: I'm kinda confused - are you assuming that all `task_url`s will fit into one `line`?

owenauch: I think that each line is just all the content until a linebreak, which in markdown should be all of the task urls even if that isn't how it appears visually

src/asana/logic.py
owenauch: Just tested it and it doesn't generate another story — seems like a no-op

src/asana/helpers.py
binoche9: Hmm... that's definitely not what I would have guessed 🤔 I'd still expect there to be a newline character (and raw editing in GitHub would also lead me to guess that way)... but if you've manually tested this and confirmed it then I have no choice but to believe you lol

owenauch: I need to verify, that was just my assumption, so I'll take a look before merging

src/asana/controller.py
```diff
@@ -33,6 +35,14 @@ def update_task(pull_request: PullRequest, task_id: str):
...
     asana_client.update_task(task_id, update_task_fields)
     asana_client.add_followers(task_id, fields["followers"])
+    maybe_complete_tasks_on_merge(pull_request)
+
+
+def maybe_complete_tasks_on_merge(pull_request: PullRequest):
+    if asana_logic.should_autocomplete_tasks_on_merge(pull_request):
+        task_ids_to_complete_on_merge = asana_helpers.get_linked_task_ids(pull_request)
+        for complete_on_merge_task_id in task_ids_to_complete_on_merge:
+            asana_client.update_task(complete_on_merge_task_id, {"completed": True})
```
gregeinfrank, line 45: nit: could be nice to just have a `asana_client.complete_task` function rather than passing in this dict of `{"completed": True}` to `update_task`

src/asana/helpers.py
```diff
@@ -309,6 +309,31 @@ def urlreplace(matchobj: Match[str]) -> str:
     return re.sub(URL_REGEX, urlreplace, text)
 
 
+def get_linked_task_ids(pull_request: PullRequest) -> List[str]:
+    """
+    Extracts linked task ids from the body of the PR.
```
gregeinfrank, line 314: The test is good for documentation, but adding a description here of the format of what we expect and what this is extracting would be helpful

src/asana/helpers.py
```diff
@@ -309,6 +309,31 @@ def urlreplace(matchobj: Match[str]) -> str:
...
+            task_url_line = line
+            break
+        if line.startswith("Asana tasks:"):
+            seen_asana_tasks_line = True
+
+    if task_url_line:
+        task_urls = task_url_line.split()
+        task_ids = []
+        for url in task_urls:
+            task_ids.append(re.findall("\d+(?!.*\d)", url)[0])
```
gregeinfrank, line 331: This is making an assumption that the task links will be in the format we expect. Can you make this code a bit safer by first running the regex, then if it matches, extract `[0]` from the result? Adding a test where the line is something like `"Asana tasks:\nfoo bar"` should be a good starting point from a TDD route

src/asana/logic.py
```diff
@@ -0,0 +1,23 @@
...
+    )
+
+
+def _is_autocomplete_feature_enabled():
+    return os.getenv("SGTM_FEATURE__AUTOCOMPLETE_ENABLED") == "true"
```
gregeinfrank, line 23: nit: I think if we're going to have these feature flags throughout the codebase, it's worth adding a module (either part of `config.py` or something similar) that handles the feature flag logic and take the env variable retrieval out of these files and isolate into one place

src/asana/logic.py
```diff
@@ -0,0 +1,23 @@
...
+from enum import Enum, unique
+import os
+
+
+@unique
+class AutocompleteLabel(Enum):
+    COMPLETE_ON_MERGE = "complete tasks on merge"
+
+
+def should_autocomplete_tasks_on_merge(pull_request: PullRequest):
```
gregeinfrank, line 12:
```suggestion
def should_autocomplete_tasks_on_merge(pull_request: PullRequest) -> bool:
```

src/asana/logic.py
```diff
@@ -0,0 +1,23 @@
+from src.github.models import PullRequest
+from src.github.helpers import pull_request_has_label
+from enum import Enum, unique
+import os
+
```
gregeinfrank, line 5: Anything that should be moved out of asanahelpers or asana controller / etc into this new `logic.py`?

src/asana/controller.py
gregeinfrank: Also - what happens if the task is already complete? Would this noop or throw an exception?

test/asana/test_controller.py
```diff
@@ -110,6 +111,35 @@ def test_updated_review_with_comments(
...
+        should_autocomplete_tasks_on_merge_mock,
+        get_linked_task_ids_mock,
+        update_task_mock,
+    ):
+        task_ids = ["123", "456"]
+        get_linked_task_ids_mock.return_value = task_ids
+        pull_request = build(builder.pull_request().merged(True))
+        controller.maybe_complete_tasks_on_merge(pull_request)
+        update_task_mock.assert_any_call(task_ids[0], {"completed": True})
+        update_task_mock.assert_any_call(task_ids[0], {"completed": True})
```
gregeinfrank, line 140:
```suggestion
        update_task_mock.assert_any_call(task_ids[1], {"completed": True})
```

gregeinfrank (changes requested):
Cool - i think the plan with `git pr` sounds good - thanks for continuing to contribute to SGTM!

src/asana/controller.py
owenauch: It's a noop

src/asana/logic.py
owenauch: Since this is already pretty big, I'd prefer to do this in a follow up PR

owenauch: https://app.asana.com/0/1149418478823393/1199932159016854/f

src/asana/helpers.py
owenauch: Verified this works properly in https://github.com/Asana/codez/pull/102360

owenauch:
I addressed all the feedback here and tested in https://github.com/Asana/codez/pull/102360, so this should be good to go. We also are fine to merge this and turn it on before the git pr change, since it will just be a noop.

src/asana/helpers.py
gregeinfrank: @owenauch  - one weird thing I realized -- the LGTM Asana user isn't in all of our projects - if you link to a non-public task, I believe SGTM won't be able to merge the task.

gregeinfrank (approved):
LGTM!

src/asana/helpers.py
binoche9: @gregeinfrank This is true for nadi as well - it won't post the PR links if the task is private, so I think this is somewhat okay

gregeinfrank: @binoche9  -- it's a bit different though because the Nadi user (https://app.asana.com/0/438505498636/list) is actually in the asana domain and has access to all non-private tasks/projects in the Asana domain, whereas LGTM (https://app.asana.com/0/783395256970426/list) is a guest with an `@corp.asana.com` domain - and by default is _not_ a member of all projects unless explicitly added to them

binoche9: Ah TIL. Thanks for the explanation.

PR #77, opened 2020-10-13
Adds PR Status description to Asana Task

arthurborem-asana:
Followed the logic in the LGTM repo:

https://github.com/Asana/LGTM/blob/6f51b17e6565b9ff203465bd4bea6b206c3f9824/src/LGTM/asana/adapter.clj#L29-L49

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1198208547880046)


arthurborem-asana:
I know this was a _long_ time ago, but this is my first offsprint since the last one!

src/asana/helpers.py
```diff
@@ -394,22 +398,26 @@ def _task_description_from_pull_request(pull_request: PullRequest) -> str:
         + "\n✍️ "
         + author
         + _generate_assignee_description(pull_request.assignee())
+        + f"\n❗️Task is {status} because {status_reason.reason}"
```
gregeinfrank, line 401: Do you think we should include the ❗ when "the pull request is open"? It seems like an alert that something is wrong, when it might not be... (I realize this was probably the original logic in the clojure code, but curious what your thoughts are... it does make the code a bit cleaner, so I think it's not the end of the world)

src/asana/helpers.py
```diff
@@ -386,6 +388,8 @@ def _task_description_from_pull_request(pull_request: PullRequest) -> str:
     author = _asana_user_url_from_github_user_handle(github_author.login())
     if author is None:
         author = _asana_display_name_for_github_user(github_author)
+    status_reason = _task_completion_from_pull_request(pull_request)
+    status = "closed" if status_reason.isClosed else "not closed"
```
gregeinfrank, line 392: nit: since we're referring to Asana tasks, should this be `"complete"` and `"incomplete"`?

src/asana/helpers.py
```diff
@@ -394,22 +398,26 @@ def _task_description_from_pull_request(pull_request: PullRequest) -> str:
         + "\n✍️ "
         + author
         + _generate_assignee_description(pull_request.assignee())
+        + f"\n❗️Task is {status} because {status_reason.reason}"
         + _wrap_in_tag("strong")("\n\nDescription:\n")
         + convert_urls_to_links(escape(pull_request.body(), quote=False))
     )
 
 
 def _task_completion_from_pull_request(pull_request: PullRequest) -> bool:
```
gregeinfrank, line 407:
```suggestion
def _task_completion_from_pull_request(pull_request: PullRequest) -> StatusReason:
```

gregeinfrank: Can you run `black` and `mypy` on the `src` and `test` directories to make sure both pass?

gregeinfrank (changes requested):
Looks good - just a few more nits / comments

src/asana/helpers.py
```diff
@@ -26,6 +26,8 @@
     "AttachmentData", "file_name file_url image_type"
 )
 
+StatusReason = collections.namedtuple("StatusReason", "isClosed reason")
```
gregeinfrank, line 29: Is it Pythonic to us `isClosed` rather than `is_closed` here?

arthurborem-asana: absolutely not! Haven't touched Python in many weeks, sorry!

src/asana/helpers.py
arthurborem-asana: I can see keeping the ❗️ as a visual indication of where to look if I want to see the status of this PR. Having it go away might cause some brief confusion regarding where to find the PR status in the body. I'm okay with either option though!

src/asana/helpers.py
gregeinfrank: @arthurborem-asana  -- I still think we should use `is_closed`

gregeinfrank: or `is_complete` even

src/asana/helpers.py
gregeinfrank: We can start with it and see if we get feedback, can always change it later

gregeinfrank (changes requested):

arthurborem-asana:
Fixed `isClosed` → `is_closed`!

gregeinfrank (approved):
LGTM!

PR #85, opened 2021-03-25
Script to set up an SGTM tasks project

gohelankit:
### Description
Based on user input, does one of:
- Create new project with the suggested custom fields
- Update existing project with the suggested custom fields and adds the PAT user as a follower to the project

### Testing
Tested for both new and existing projects 

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1200111698470481)

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Again, you will probably want to create a new Github user in your org that is ju
 1. For the Github user you want to use, generate a [Personal Access Token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) with the following permissions:
    * repo (Full control of private repositories)
    * read:org (Read org and team membership, read org projects)
-1. Generate a [secret token](https://developer.github.com/webhooks/securing/) for your Github webhook. Github suggests generating this via `ruby -rsecurerandom -e 'puts SecureRandom.hex(20)'`, but use whatever method you are comfortable with to generate a secure secret token. Save this somewhere, as you'll need it twice in the later steps.
+2. Generate a [secret token](https://developer.github.com/webhooks/securing/) for your Github webhook. Github suggests generating this via `ruby -rsecurerandom -e 'puts SecureRandom.hex(20)'`, but use whatever method you are comfortable with to generate a secure secret token. Save this somewhere, as you'll need it twice in the later steps.
 
 Copy this Personal Accesss Token for the next step.
 
@@ -44,14 +44,22 @@ Copy this Personal Accesss Token for the next step.
 You'll need to create two Asana projects: one that will store the mapping of Github username to Asana user id, and the other where your Github sync tasks will live.
 
 1. Create your "SGTM Users" project (feel free to name this whatever you want -- this is just a suggestion). The requirements of this project are two custom fields named: "Github Username" (Text field) and "user_id" (Number field). Save the `id` of this project (from the URL once created) in `./terraform/terraform.tfvars.json` under `"asana_users_project_id"`.
-1. Create your "SGTM <repo> tasks" project (feel free to name this whatever you want -- this is just a suggestion). Suggested custom fields (though optional) are: "PR Status" (with values: "Open", "Merged", and "Closed") and "Build" (with values: "Success", "Failure", "Error"). When these exist on your project, SGTM will update these custom fields based on the merge status and build status of your pull requests.
-   1. If you have multiple repositories you want synced to Asana, you can create several of these projects. Make sure to take note of all of the project IDs for a later step.
-1. Make sure that the Asana user/guest that you created earlier is a member of both of these projects.
+2. To create your "SGTM <repo> tasks" project, use the `setup_sgtm_tasks_project.py` script. The script will prompt you for the PAT you generated earlier, and guide you through setting up a brand new project or updating an existing project with the recommended Custom Fields.
+     ```
+      >>> To setup a new project
+      python3 scripts/setup_sgtm_tasks_project.py  -p "<PAT>" create -n "<PROJECT NAME>" -t "<TEAM ID>"
+           
+      >>> To update an existing project with the suggested custom fields
+      python3 scripts/setup_sgtm_tasks_project.py  -p "<PAT>" update -e "<EXISTING PROJECT ID>"
+      ```
+    1. If you have multiple repositories you want synced to Asana, you can create several of these projects. Make sure to take note of all of the project IDs for a later step.
+    2. If you are on Asana Basic and do not have access to Custom Fields, the script will skip that step - SGTM will work even without the suggested fields
+3. Make sure that the Asana user/guest that you created earlier is a member of both of these projects.
 
 ### Set your Terraform variables
 NOTE: AWS S3 Bucket names are globally unique, so you will need to choose your own bucket names to be unique that no other AWS account has already created.
 1. In `./terraform/variables.tf`, any variable that is listed without a default value needs to be set. The preferred method of setting these values is through [environment variables](https://www.terraform.io/docs/cli/config/environment-variables.html#tf_var_name). For example, to se terraform variable `asana_users_project_id`, you'll want to set an environment variable `TF_VAR_asana_users_project_id`.
-1. Save these somewhere that you and others collaborating on this deployment could share (we save ours in an Asana task internally, of course) since these will need to be the same each time you apply new changes.
+2. Save these somewhere that you and others collaborating on this deployment could share (we save ours in an Asana task internally, of course) since these will need to be the same each time you apply new changes.
 
 ### Zip up your code
 From the root of your repository directory, run `./scripts/zip_lambda_code.sh`. This will zip up all of the Python code and dependencies to be pushed to AWS in the next step (the `terraform apply`)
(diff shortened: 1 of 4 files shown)


scripts/setup.py
```diff
@@ -92,7 +92,8 @@ def setup_state(args):
 
     s3_client = boto3.client("s3", region_name=REGION)
     s3_client.create_bucket(
-        ACL="private", Bucket=bucket_name,
+        ACL="private",
```
gohelankit, line 95: No changes to this file, I ran `black` on the scripts directory and it also formatted this

scripts/setup_sgtm_tasks_project.py
```diff
@@ -0,0 +1,196 @@
...
+    client.add_user_to_project(project_id)
+
+
+class AsanaClient(object):
+    def __init__(self, personal_access_token: str):
+        self.client = asana.Client.access_token(personal_access_token)
+        self.client.headers = {"Asana-Enable": "string_ids"}
+        self.workspace_id = self._get_workspace_id()
+
+    def _get_workspace_id(self) -> str:
```
gohelankit, line 107: I'm not fully sold on this so I'm curious to get your opinion - the other option would be to explicitly take in a workspace ID (but that might not be trivial for the user?)

scripts/setup_sgtm_tasks_project.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+if __name__ == "__main__":
+    args = parse_args()
+
+    if args.new_project_name and args.existing_project_id:
+        print(
+            "Cannot select both an existing project and a new project, please try again"
+        )
+        sys.exit(0)
```
gregeinfrank, line 183: I'd consider leaning into `argparse` to have two subcommands: `update` or `create`, then you wouldn't have to manually check for conflicting options

scripts/setup_sgtm_tasks_project.py
gregeinfrank: This is just picking the first in the list -- what about a more interactive version if there are multiple workspaces and having the user choose from the available workspace ids?

scripts/setup_sgtm_tasks_project.py
```diff
@@ -0,0 +1,197 @@
...
+        "--new_project_name",
+        help="""The Asana Project name - use this if you want to set up a project from scratch
+		""",
+        type=str,
+        required=False,
+    )
+
+    parser.add_argument(
+        "-t",
+        "--team_id",
```
gregeinfrank, line 69: What's a good way for users to figure out this ID?

scripts/setup_sgtm_tasks_project.py
```diff
@@ -0,0 +1,197 @@
...
+        description="""A helper script to set up an SGTM tasks Asana project with the suggested custom fields
+		"""
+    )
+
+    parser.add_argument(
+        "-p",
+        "--personal_access_token",
+        help="""The Asana Personal Access Token that will be used to set up the Asana project. If you do not already have one,
+		follow the instructions at https://developers.asana.com/docs/personal-access-token
+		""",
```
gregeinfrank, line 53: Should there be some documentation why this would be a different access token than SGTM will actually use?

scripts/setup_sgtm_tasks_project.py
gohelankit: Nice, TIL!

scripts/setup_sgtm_tasks_project.py
gohelankit: If you click on the team page in Asana, you can get it from the URL. Should I add that to the help?

scripts/setup_sgtm_tasks_project.py
gohelankit: I'm not sure what you mean - can you elaborate?

scripts/setup_sgtm_tasks_project.py
gohelankit: Good point - `fzf` might be useful here

gregeinfrank: Would be nice, but I'd prefer not introducing more external dependencies to folks' setup if not necessary

scripts/setup_sgtm_tasks_project.py
gregeinfrank: In the setup of the project itself you need to specify an access token for the user who will be creating the tasks from SGTM. This could be the same or a different access token than the one they'd be passing in here. For example - our SGTM user right now at Asana is a guest, but I (@gregeinfrank) created the projects

scripts/setup_sgtm_tasks_project.py
gohelankit: Is the alternative to manually write a function that provides a CLI menu sort of thing? That's doable but feels like reinventing the wheel

scripts/setup_sgtm_tasks_project.py
gohelankit: How about
```
The PAT you pass in here can be the same/different from the PAT created in the "Create your credentials" section
```
?

gohelankit (commented):
@gregeinfrank I addressed most of your comments, apart from the interactive workspace ID thing

scripts/setup_sgtm_tasks_project.py
gregeinfrank: Yeah the alternative is to print out a numbered list and ask for input (1-N)

scripts/setup_sgtm_tasks_project.py
gregeinfrank: Sure -- and do we want to fall back to the PAT in the environment variables if not passed in?

scripts/setup_sgtm_tasks_project.py
```diff
@@ -0,0 +1,193 @@
...
+    color: str
+
+
+@dataclass
+class CustomField:
+    name: str
+    enum_options: List[EnumOption]
+
+
+CUSTOM_FIELDS = [
```
gregeinfrank, line 22: I just thought of something -- is it possible for customers to have access to the API but not custom fields? Is it safe to assume that anyone using SGTM is on a non-free tier? I'll follow-up with the API team about this.

gregeinfrank: OK - just confirmed https://asana.slack.com/archives/C0PHY07V2/p1616773926047500 -- I think we need to cover the case where a domain doesn't have access to custom fields. SGTM will still work with projects that don't have the custom fields, so we should be able to detect this somehow and not attempt to add the custom fields. Happy to pair on ideas here

gohelankit: Ah that's a good point - I didn't think of that. I followed up on the thread and Kevin mentioned that there isn't a way to get premiumness via the API so we might need to think of an alternate solution.

One thing we could do is try to create this irrespective of premiumness, and I assume that the API should return an error if the operation isn't possible. I've asked the API team if this is true

gohelankit: Response:
> You'll get a 402 status code with Custom Fields are not available for free users or guests as the message

gohelankit: Slightly concerned about the guest part here based on the instructions in our setup

scripts/setup_sgtm_tasks_project.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def _get_workspace_id(self) -> str:
+        """
+        Get the workspace ID that the user who created the PAT is a part of
+        """
+        pat_user = self.client.users.me()
+        if len(pat_user["workspaces"]) == 1:
+            return pat_user["workspaces"][0]["gid"]
+        else:
+            workspaces = pat_user["workspaces"]
```
gohelankit, line 132: Tested this manually in the python console by setting up some fake data - I wasn't sure how else to test this

scripts/setup_sgtm_tasks_project.py
```diff
@@ -0,0 +1,228 @@
...
+                        "color": enum_option.color,
+                        "enabled": True,
+                    }
+                )
+
+            try:
+                response = self.client.custom_fields.create_custom_field(
+                    custom_field_data
+                )
+            except asana.error.PremiumOnlyError:
```
gohelankit, line 199: Tested this by setting up a free user through the NUD and generating a PAT for that user

gregeinfrank (approved):
Nice-  thanks for the work here. I think this looks good - the last thing to do would be to update the `README` with the instructions on how to set up the project with this script

README.md
```diff
@@ -60,7 +68,7 @@ From the root of your repository directory, run `./scripts/zip_lambda_code.sh`.
 You'll first need to set up the [Terraform remote state](https://www.terraform.io/docs/state/remote.html) to be the source of truth for the state of your deployed infrastructure.
 
 1. Run `python3 ./scripts/setup.py state` (this will create  an S3 bucket and DyanmoDb lock table for Terraform)
-1. Initialize and apply the infrastructure:
+2. Initialize and apply the infrastructure:
```
gregeinfrank, line 71: @gohelankit  FYI - I like to keep this all numbered with `1.` because Markdown will parse it into an actual numbered list for you, and it makes adding / reordering the lines super easy

PR #86, opened 2021-03-25
Comment on pull request when auto-merge didn't merge due to a merge conflict

maryliang8:
Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1200113720472464)

### DIFF
--- a/src/github/logic.py
+++ b/src/github/logic.py
@@ -169,7 +169,9 @@ def maybe_add_automerge_warning_comment(pull_request: PullRequest):
             and pull_request_has_label(
                 pull_request, AutomergeLabel.AFTER_TESTS_AND_APPROVAL.value
             )
-            and not _pull_request_has_automerge_comment(pull_request)
+            and not _pull_request_has_warning_comment(
+                pull_request, AUTOMERGE_COMMENT_WARNING
+            )
             and not pull_request.is_approved()
         ):
 
@@ -221,6 +223,9 @@ def _is_pull_request_ready_for_automerge(pull_request: PullRequest) -> bool:
         if (
             pull_request.is_build_successful()
             and pull_request.mergeable() == MergeableState.CONFLICTING
+            and not _pull_request_has_warning_comment(
+                pull_request, AUTOMERGE_CONFLICT_COMMENT_WARNING
+            )
         ):
             add_automerge_conflict_failure_comment(pull_request)
 
@@ -233,6 +238,9 @@ def _is_pull_request_ready_for_automerge(pull_request: PullRequest) -> bool:
             pull_request.is_build_successful()
             and pull_request.is_approved()
             and pull_request.mergeable() == MergeableState.CONFLICTING
+            and not _pull_request_has_warning_comment(
+                pull_request, AUTOMERGE_CONFLICT_COMMENT_WARNING
+            )
         ):
             add_automerge_conflict_failure_comment(pull_request)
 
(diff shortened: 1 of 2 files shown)


src/github/logic.py
```diff
@@ -217,20 +218,42 @@ def _is_pull_request_ready_for_automerge(pull_request: PullRequest) -> bool:
         )
 
     if pull_request_has_label(pull_request, AutomergeLabel.AFTER_TESTS.value):
```
gregeinfrank, line 220: I think we would run into a possible issue of this comment getting added over and over again in a loop and probably need a way to prevent that

gregeinfrank (changes requested):

src/github/logic.py
maryliang8: We could do a check to see if the comment already exists!

gregeinfrank: sounds good! Looks like you have done this recently and should be pretty straightforward :)

test/github/test_logic.py
```diff
@@ -49,6 +49,82 @@ def test_handle_status_webhook_not_ready_for_automerge(
...
+            )
+        )
+
+        merged = github_logic._is_pull_request_ready_for_automerge(pull_request)
+
+        self.assertFalse(merged)
+        merge_pull_request_mock.assert_not_called()
+        add_pr_comment_mock.assert_not_called()
+
+    def test_does_not_add_comment_not_ready_for_automerge_due_to_conflict_twice(
```
gregeinfrank, line 92: I think the test names are switched here -- this test is testing that a comment _is_ added, right?

test/github/test_logic.py
```diff
@@ -49,6 +49,82 @@ def test_handle_status_webhook_not_ready_for_automerge(
...
+            .comments(
+                [
+                    builder.comment()
+                    .author(builder.user("github_unknown_user_login"))
+                    .body(github_logic.AUTOMERGE_CONFLICT_COMMENT_WARNING)
+                ]
+            )
+            .label(
+                builder.label().name(
+                    github_logic.AutomergeLabel.AFTER_TESTS_AND_APPROVAL.value
```
gregeinfrank, line 81: These tests are only testing for `AFTER_TESTS_AND_APPROVAL`, but there are two codepaths that add the comment (the other is for `AFTER_TESTS`) -- is there a way to test both without completely re-writing both of these tests?

test/github/test_logic.py
```diff
@@ -49,6 +49,82 @@ def test_handle_status_webhook_not_ready_for_automerge(
...
+                    builder.review()
+                    .submitted_at("2020-01-13T14:59:57Z")
+                    .state(ReviewState.APPROVED)
+                ]
+            )
+            .comments(
+                [
+                    builder.comment()
+                    .author(builder.user("github_unknown_user_login"))
+                    .body(github_logic.AUTOMERGE_CONFLICT_COMMENT_WARNING)
```
gregeinfrank, line 76: maybe leave a comment in the code here explaining that this comment is meant to test the case where a comment already exists for the automerge warning, and that you're testing that another comment isn't added

test/github/test_logic.py
```diff
@@ -49,6 +49,82 @@ def test_handle_status_webhook_not_ready_for_automerge(
...
+                ]
+            )
+            .label(
+                builder.label().name(
+                    github_logic.AutomergeLabel.AFTER_TESTS_AND_APPROVAL.value
+                )
+            )
+        )
+
+        merged = github_logic._is_pull_request_ready_for_automerge(pull_request)
```
gregeinfrank, line 86: nit: now reading this in the tests, it's a bit surprising that a function called `_is_pull_request_ready_for_automerge` has a side effect (adds a comment). Is there another pattern or name that we can use to indicate that?

gregeinfrank (changes requested):
Looks good! Just a few comments about the tests and naming

PR #87, opened 2021-03-25
Circleci project setup

gregeinfrank:
Add two jobs: test and black, black should only run on non-master branches

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1200114318401080)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,17 @@ orbs:
   python: circleci/python@0.2.1
 
 jobs:
+  black:
+    executor: python/default
+    steps:
+      - checkout
+      - python/load-cache
+      - python/install-deps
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,40 @@
...
+      - run: pip install -r requirements-dev.txt
+      - python/save-cache
+
+jobs:
+  black:
+    executor: python/default
+    steps:
+      - setup
+      - run:
+          command: black --check src test scripts
```
gregeinfrank, line 21: I'd like to make this step fix the files and commit it back to the branch, but I'll attempt that in a follow-up

gohelankit (approved):
Haven't set this up before but the yaml config looks reasonable to me

PR #100, opened 2021-08-03
Add a Markdown parser to fix bugs with comment parsing

owenauch:
I got tired of trying to implement this one off, and decided to add an extensible markdown parser. This should get us most of the way there, but I'm having some issues with getting task descriptions to look right — it seems like the parser doesn't add line breaks for each time we return in markdown, which is correct according to the spec but makes our task descriptions look weird, so we might need to change this via git pr or something.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1200710613461122)


src/asana/helpers.py
```diff
@@ -294,20 +291,6 @@ def create_attachments(body_text: str, task_id: str) -> None:
 }
 
 
-def convert_urls_to_links(text: str) -> str:
```
owenauch, line 297: this is now as part of the markdown parser

test/asana/helpers/test_asana_comment_from_github_comment.py
```diff
@@ -34,7 +34,7 @@ def test_transforms_urls_from_comment_tect(self):
             github_comment
         )
         self.assertContainsStrings(
-            asana_comment, ['<A href="{}">{}</A>'.format(escape(url), url)]
+            asana_comment, ['<a href="{}">{}</a>'.format(escape(url), url)]
```
owenauch, line 37: changed for consistency

owenauch:
See https://app.asana.com/0/1127984365697766/1200318006155092/f and https://github.com/Asana/codez/pull/110935 for an example of the description problem.

src/markdown_parser.py
```diff
@@ -0,0 +1,38 @@
...
+        return text + "\n"
+
+    def block_quote(self, text):
+        return f"<em>> {escape(text, quote=False)}</em>"
+
+    def strikethrough(self, text):
+        return f"<s>{escape(text, quote=False)}</s>"
+
+    def heading(self, text, level):
+        return f"\n<b>{text}</b>\n"
```
gregeinfrank, line 21: Should this be `escape`d as well?

test/test_markdown_parser.py
```diff
@@ -0,0 +1,58 @@
+import unittest
+
+from src.markdown_parser import convert_github_markdown_to_asana_xml
+
+
+class TestConvertGithubMarkdownToAsanaXml(unittest.TestCase):
+    def test_basic_markdown(self):
+        md = """~~strike~~ **bold** _italic_ `code` [link](asana.com)"""
```
gregeinfrank, line 8: Have you tested a couple of the examples in https://app.asana.com/0/780306770840675/1200376931872355/f that broke the old version?

gregeinfrank (commented):
Task description in https://app.asana.com/0/1127984365697766/1200318006155092/f looks OK to me -- what is the issue you're seeing?

owenauch:
> Task description in https://app.asana.com/0/1127984365697766/1200318006155092/f looks OK to me -- what is the issue you're seeing?

The spacing between different sections feels weird and harder to read but I think that might just be an issue of how we format the initial description message rather than the parser

src/markdown_parser.py
owenauch: Yes, good catch

test/test_markdown_parser.py
owenauch: Good call, I'll do that before merging

owenauch: Tested this on most of the comments and added one change to handle the fact that the Asana API can't handle <pre> tags

src/markdown_parser.py
```diff
@@ -0,0 +1,43 @@
...
+def convert_github_markdown_to_asana_xml(text: str) -> str:
+    markdown = mistune.create_markdown(
+        renderer=GithubToAsanaRenderer(escape=False), plugins=["strikethrough"],
+    )
+    return _strip_pre_tags(markdown(text))
+
+
+# the Asana API doesn't accept pre tags so we strip them
+def _strip_pre_tags(text: str) -> str:
+    return text.replace("<pre>", "").replace("</pre>", "")
```
owenauch, line 43: I couldn't find a way to do this more cleanly through the mistune API so I gave up and did this

gregeinfrank: What happens if you implement a function like this:

```python
def block_code(self, code, lang):
    return escape(code, quote=False)
````

File: test/asana/helpers/test_extract_task_fields_from_pull_request.py [hunk 6]
```diff
@@ -44,9 +44,6 @@ def test_html_body(self):
             '<A href="https://foo.bar/baz">https://foo.bar/baz</A>',
             "✍",
             "TEST_USER_ASANA_DOMAIN_USER_ID",
-            "<strong>",
-            "Description:",
```
gregeinfrank, line 48: Why was "Description" removed?

gregeinfrank (approved):
LGTM - but just two remaining questions that aren't blocking

test/asana/helpers/test_extract_task_fields_from_pull_request.py
owenauch: It felt redundant once Summary was bolded, but I don't feel strongly about that at all

src/markdown_parser.py
owenauch: I guess I'm concerned there's some other case that will break because of `pre` so I wanted something more general, but that would work in this case

test/asana/helpers/test_extract_task_fields_from_pull_request.py
gregeinfrank: I think "Summary" is just in our CODEZ `git pr` template, but SGTM is used for other repos (and is open sourced now!) so we shouldn't necessarily optimize for the `git pr` template

owenauch: Ah makes sense, I'll add it back

owenauch:
```
test all the new functionality
```

> block quote

src/markdown_parser.py
```diff
@@ -0,0 +1,43 @@
...
+# https://gist.github.com/gruber/8891611
+URL_REGEX = r"""(?i)([^"\>\<\/\.]|^)\b((?:https?:(/{1,3}))(?:[^\s()<>{}\[\]]+|\([^\s()]*?\([^\s()]+\)[^\s()]*?\)|\([^\s]+?\))+(?:\([^\s()]*?\([^\s()]+\)[^\s()]*?\)|\([^\s]+?\)|[^\s`!()\[\]{};:'".,<>?«»“”‘’]))"""
+
+
+class GithubToAsanaRenderer(mistune.HTMLRenderer):
+    def paragraph(self, text):
+        return text + "\n"
+
+    def block_quote(self, text):
+        return f"<em>> {escape(text, quote=False)}</em>"
```
gregeinfrank, line 16: @owenauch  - I think this `<em>>` is causing SGTM to break - I assume this extra `>` was not intentional right?

see: https://app.asana.com/0/0/1200718797414710/1200725140636985/f

gregeinfrank: Nevermind -- I think https://github.com/Asana/SGTM/pull/101/files should fix it

PR #101, opened 2021-08-04
Escape extra bracket in block quote

gregeinfrank:
Adding in an extra `>` will cause the Asana html parser to break -- this should be escaped

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1200725692785455)

### DIFF
--- a/src/markdown_parser.py
+++ b/src/markdown_parser.py
@@ -13,7 +13,7 @@ class GithubToAsanaRenderer(mistune.HTMLRenderer):
         return text + "\n"
 
     def block_quote(self, text):
-        return f"<em>> {escape(text, quote=False)}</em>"
+        return f"<em>{escape('> ' + text, quote=False)}</em>"
 
     def strikethrough(self, text):
         return f"<s>{escape(text, quote=False)}</s>"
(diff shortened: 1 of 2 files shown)


owenauch:
Ah yes thank you, though I'm surprised this didn't break when I manually tested this

owenauch (approved):

gregeinfrank:
Do you think you can test this one manually if you already have a PR set up for local testing?

owenauch:
Yeah I can test this one

owenauch:
Oh img tags also make it sad. We automate attachment extraction after the fact, so I think we can just remove img tags from the comment body?

gregeinfrank:
@owenauch can you point me to the test PR you're using?

owenauch:
This is the test comment: https://app.asana.com/0/0/1200318006155092/1200330488149808/f — I've been just changing this comment and rerunning the same command.

I added some logic to strip raw html tags that are added to the comment, which would solve the img tag problem, but this comment is still broken, I think because of the nested block quote. I'm not sure how concerned to be about that specific edge case.

owenauch:
I can push that logic up if you want, but let me know

gregeinfrank:
Interesting.. sure - feel free to push to this branch

PR #104, opened 2021-09-15
Fallback to statusCheckRollup when status is none

gregeinfrank:
On some repos, we're seeing [this issue](https://github.community/t/graphql-api-response-for-commit-status-is-null/13762/6) - so falling back to statusCheckRollup seems to do the trick

Fixes: https://app.asana.com/0/1149418478823393/1200751774126749/f

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1200991024698718)


stevelandeyasana (approved):
Thanks!

PR #106, opened 2021-09-23
Improve markdown parsing

gregeinfrank:
Fix a few reported bugs with nested markdown parsing (see: https://app.asana.com/0/780306770840675/1200944119020029/f and https://app.asana.com/0/780306770840675/1200842082019847/f)

Biggest change here is to no longer blindly strip all `<` and `>`, and add an `inline_html` function as a catch-all for these and escape the html.

Added some test cases from the reported bugs

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1201049190615122)

### DIFF
--- a/src/markdown_parser.py
+++ b/src/markdown_parser.py
@@ -13,17 +13,23 @@ class GithubToAsanaRenderer(mistune.HTMLRenderer):
         return text + "\n"
 
     def block_quote(self, text):
-        return f"<em>{escape('> ' + text, quote=False)}</em>"
+        return f"<em>&gt; {text}</em>"
 
     def strikethrough(self, text):
-        return f"<s>{escape(text, quote=False)}</s>"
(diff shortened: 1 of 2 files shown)


src/markdown_parser.py
```diff
@@ -9,23 +9,29 @@
...
+        return f"<s>{text}</s>"
 
-    def heading(self, text, level):
-        return f"\n<b>{escape(text, quote=False)}</b>\n"
+    def heading(self, text, level) -> str:
+        return f"\n<b>{text}</b>\n"
 
-    def thematic_break(self):
+    def thematic_break(self) -> str:
         # Asana API doesn't not support <hr />
```
owenauch, line 25: nit: this was my very confusing typo from before

owenauch (approved):
Thank you for fixing this — sorry this halfway done state has been noisy

PR #110, opened 2021-11-11
Create review subtasks for requested reviewers

magnus-asana:
This is still WIP.

### DIFF
--- a/src/asana/client.py
+++ b/src/asana/client.py
@@ -59,6 +59,31 @@ class AsanaClient(object):
         response = self.asana_api_client.tasks.create(create_task_params)
         return response["gid"]
 
+    def create_subtask(
+        self, parent_task_id: str, reviewer: str, task_name: str, task_description,
+        due_date_str: str = None
+    ) -> str:
+        """
+        Creates an Asana subtask for the given parent task, returning the task_id
+        """
+        validate_object_id(parent_task_id, "AsanaClient.create_subtask requires a task_id")
+
+        create_task_params = {
+            "name": task_name,
+            "html_notes": task_description,
+            "assignee": reviewer,
+            "parent": parent_task_id
+        }
+        if due_date_str:
+            create_task_params["due_on"] = due_date_str
+        response = self.asana_api_client.tasks.create(create_task_params)
+        return response["gid"]
+
+    def get_task_completed_status(self, task_id: str) -> bool:
+        """Returns bool value representing the task completed status."""
+        response = self.asana_api_client.tasks.find_by_id(task_id, opt_fields=["completed"])
+        return response["completed"]
+
     def update_task(self, task_id: str, fields: dict):
         """
         Updates the specified Asana task, setting the provided fields
@@ -146,11 +171,27 @@ def create_task(project_id: str, due_date_str: str = None) -> str:
     return AsanaClient.singleton().create_task(project_id, due_date_str=due_date_str)
 
 
+def create_subtask(
+    parent_task_id: str, reviewer_id: str, task_name: str, task_description,
+    due_date_str: str = None
+) -> str:
+    """
+    Creates an Asana task and makes is a subtask of the given task id
+    """
+    return AsanaClient.singleton().create_subtask(
+        parent_task_id, reviewer_id, task_name, task_description, due_date_str
+    )
+
+
 def update_task(task_id: str, fields: dict):
     """
     Updates the specified Asana task, setting the provided fields
     """
     return AsanaClient.singleton().update_task(task_id, fields)
+
+
+def is_task_completed(task_id) -> bool:
+    return AsanaClient.singleton().get_task_completed_status(task_id)
 
 
 def complete_task(task_id: str):
--- a/src/asana/controller.py
+++ b/src/asana/controller.py
@@ -38,6 +38,29 @@ def update_task(pull_request: PullRequest, task_id: str):
     maybe_complete_tasks_on_merge(pull_request)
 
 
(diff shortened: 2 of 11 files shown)


src/asana/client.py
```diff
@@ -59,6 +59,31 @@ def create_task(self, project_id: str, due_date_str: str = None) -> str:
         response = self.asana_api_client.tasks.create(create_task_params)
         return response["gid"]
 
+    def create_subtask(
+        self, parent_task_id: str, reviewer: str, task_name: str, task_description,
```
gregeinfrank, line 63: nit: I'd just call this `assignee` since this is a wrapper around the Asana API and doesn't necessarily need to know the context of "reviewer" - just to format and forward the request to the API

```suggestion
        self, parent_task_id: str, assignee: str, task_name: str, task_description,
```

src/asana/client.py
```diff
@@ -146,13 +171,29 @@ def create_task(project_id: str, due_date_str: str = None) -> str:
...
+
+
 def update_task(task_id: str, fields: dict):
     """
     Updates the specified Asana task, setting the provided fields
     """
     return AsanaClient.singleton().update_task(task_id, fields)
 
 
+def is_task_completed(task_id) -> bool:
```
gregeinfrank, line 193: I think the pattern established here is to keep the naming consistent

```suggestion
def get_task_completed_status(task_id) -> bool:
```

src/asana/controller.py
```diff
@@ -38,6 +38,29 @@ def update_task(pull_request: PullRequest, task_id: str):
     maybe_complete_tasks_on_merge(pull_request)
 
 
+def create_review_subtask(pull_request: PullRequest, parent_task_id: str, reviewer_handle: str):
```
gregeinfrank, line 41:
```suggestion
def create_review_subtask(pull_request: PullRequest, parent_task_id: str, reviewer_handle: str) -> str:
```

src/asana/controller.py
```diff
@@ -38,6 +38,29 @@ def update_task(pull_request: PullRequest, task_id: str):
...
+    task_description = asana_helpers.subtask_description_from_pull_request(
+        pull_request, reviewer_handle
+    )
+    due_date_str = asana_helpers.default_due_date_str()
+    return asana_client.create_subtask(
+        parent_task_id, asana_id, task_name, task_description, due_date_str=due_date_str
+    )
+
+
+def update_subtask(pull_request: PullRequest, subtask_id):
```
gregeinfrank, line 53: The naming here is a bit confusing, perhaps just `reopen_subtask_if_completed` would be more accurate?

src/dynamodb/client.py
```diff
@@ -210,6 +210,33 @@ def get_asana_domain_user_id_from_github_handle(github_handle: str) -> Optional[
...
+
+def insert_two_github_node_to_asana_id_mapping(gh_node_id_a: str, gh_node_id_b: str, asana_id: str):
+    """
+        Using the singleton instance of DynamoDbClient, creating it if necessary:
+
+        Creates an association between two GitHub node-ids and an Asana object-id. The dynamoDb
+        key is formed by concatenating the two GitHub node ids using a "-" separator.
+    """
+    dynamo_db_key = _get_dynamodb_key_from_two_github_nodes(gh_node_id_a, gh_node_id_b)
+    print(f"Inserting key {dynamo_db_key} to DynamoDb.")
```
gregeinfrank, line 234:
```suggestion
    logger.info(f"Inserting key {dynamo_db_key} to DynamoDb.")
```

src/dynamodb/client.py
```diff
@@ -236,3 +263,6 @@ def bulk_insert_github_handle_to_asana_user_id_mapping(
     DynamoDbClient.singleton().bulk_insert_github_handle_to_asana_user_id_mapping(
         gh_and_asana_ids
     )
+
+def _get_dynamodb_key_from_two_github_nodes(gh_node_id_a: str, gh_node_id_b: str) -> str:
+    return f"{gh_node_id_a}-{gh_node_id_b}"
```
gregeinfrank, line 268: Perhaps this function should sort the two node ids, so we don't have to remember which order to pass them in and reduce bugs around unexpected orders

src/github/controller.py
```diff
@@ -42,6 +45,63 @@ def _add_asana_task_to_pull_request(pull_request: PullRequest, task_id: str):
...
+
+            logger.info(
+                f"Subtask created for pull request {pull_request.id()}: {task_id}: {created_subtask_id}"
+            )
+            dynamodb_client.insert_two_github_node_to_asana_id_mapping(
+                pull_request.id(), reviewer.id(), created_subtask_id
+            )
+
+
+def update_subtasks(pull_request: PullRequest, task_id: str) -> None:
```
gregeinfrank, line 80: I think it'd be more efficient to do this and `upsert_subtasks` in one loop rather than two separate functions. Mostly because it would cut the calls to dynamodb in half and updating wouldn't be necessary for newly created subtasks

gregeinfrank (commented):
Awesome start - thanks so much for picking this up. A couple comments, but directionally this looks good. Of course before getting ready for real review we'll need to add tests.

I'm also a _bit_ worried about how this rollout will work at Asana, and anticipate lots of complaints from engineers who are going to get duplicate tasks. That's probably OK as long as we get ahead of it and communicate it in advance and get people to change their behavior to use CC correctly and not add everyone as reviewers no matter what.

Also good to loop in the code review AOR holder before rolling out, I think they were involved in the original discussion around this but this is a big enough process change that there should be a coordinated rollout. Thanks!

skeggse:
It'd be so nice to have something like this!

PR #112, opened 2022-03-03
Add support for check_suite webhook

gregeinfrank:
Add support for the `check_suite` webhook action, which should have a similar effect as the `status` webhook action, but the GitHub payload is slightly different

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1201915596131251)


owenauch (approved):
LGTM!

PR #114, opened 2022-03-28
remove wrapping of inline comments in <ul>

gregeinfrank:
This was causing parsing issues for inline html within an inline comment - and added unnecessary complexity for not much benefit. This removes the `<ul>` list of inline comments for a GH comment and separates them out by newlines instead.

fixes: https://app.asana.com/0/1149418478823393/1201442536592555/f

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1202038153443517)

### DIFF
--- a/src/asana/helpers.py
+++ b/src/asana/helpers.py
@@ -350,14 +350,12 @@ def asana_comment_from_github_review(review: Review) -> str:
 
     # For each comment, prefix its text with a bracketed number that is a link to the Github comment.
     inline_comments = [
-        _wrap_in_tag("li")(
             _wrap_in_tag("A", attrs={"href": comment.url()})(f"[{i}] ")
             + _format_github_text_for_asana(comment.body())
-        )
         for i, comment in enumerate(review.comments(), start=1)
     ]
     if inline_comments:
-        comments_html = _wrap_in_tag("ul")("".join(inline_comments))
+        comments_html = "\n\n".join(inline_comments)
         if not review.is_just_comments():
             # If this was an inline reply, we already added "and left inline comments" above.
             comments_html = (
--- a/src/asana/helpers.py
+++ b/src/asana/helpers.py
@@ -355,7 +355,7 @@ def asana_comment_from_github_review(review: Review) -> str:
         for i, comment in enumerate(review.comments(), start=1)
     ]
     if inline_comments:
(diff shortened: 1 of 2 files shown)


src/markdown_parser.py
```diff
@@ -31,6 +31,9 @@ def inline_html(self, html) -> str:
     def block_html(self, html) -> str:
         return escape(html)
 
+    def linebreak(self) -> str:
```
ericwalton-asana, line 34: Looks like this is unused. Did you mean to call it in line 358 above?

src/asana/helpers.py
```diff
@@ -350,14 +350,12 @@ def asana_comment_from_github_review(review: Review) -> str:
...
-            _wrap_in_tag("A", attrs={"href": comment.url()})(f"[{i}] ")
-            + _format_github_text_for_asana(comment.body())
-        )
+        _wrap_in_tag("A", attrs={"href": comment.url()})(f"[{i}] ")
+        + _format_github_text_for_asana(comment.body())
         for i, comment in enumerate(review.comments(), start=1)
     ]
     if inline_comments:
-        comments_html = _wrap_in_tag("ul")("".join(inline_comments))
+        comments_html = "\n".join(inline_comments)
```
ericwalton-asana, line 358: I think I'd recommend two new lines, so that there's a blank space between comments.

ericwalton-asana (changes requested):
Two little nits. Seems very reasonable to do.

src/markdown_parser.py
gregeinfrank: This overrides the default from [mistune](https://github.com/lepture/mistune/blob/46a65a978b005651dcc54a340e44aef71f9f7e95/mistune/renderers.py#L163) and is used internally by that parsing library

src/asana/helpers.py
gregeinfrank: I originally had it that way, but when I went to test it, it ended up with 3 newlines (this ends up with 2). I think this has to do with the fact that each inline comment itself ends in a `"\n"`, so this essentially adds the blank space for us

gregeinfrank:
I will update `black` version in a follow-up PR, there seems to be an issue running it in CI

ericwalton-asana (approved):
LGTM!

PR #116, opened 2022-03-29
add "Draft" as PR Status option

lance-rutkin:
This PR:
- pulls in and exposes the `isDraft` field from the GitHub graphql PR object
- uses said field to determine if the "Draft" option of the "PR Status" field should be selected
- creates "Draft" as an option in the setup script
- No changes need to be made to our automerging logic as a PR being a draft is already included in a PR not being mergable

Consideration was given to adding "Draft" as an option to the "PR Status" field if, when attempting to select this option, it was not present. Ultimately, I felt this added too much complexity to the code and introduced unnecessary mingling of concerns. Instead, we will fail silently to set the field's option. We can add the option manually in Asana's SGTM projects. 

Projects that do not have the "Draft" option should still have the benefit of seeing the custom field change from nothing to "Open" when a PR is changed to or from a draft.

### DIFF
--- a/scripts/setup_sgtm_tasks_project.py
+++ b/scripts/setup_sgtm_tasks_project.py
@@ -25,6 +25,7 @@ CUSTOM_FIELDS = [
         name="PR Status",
         enum_options=[
             EnumOption(name="Open", color="green"),
+            EnumOption(name="Draft", color="grey"),
             EnumOption(name="Merged", color="purple"),
             EnumOption(name="Closed", color="red"),
         ],
--- a/src/asana/helpers.py
+++ b/src/asana/helpers.py
@@ -75,7 +75,9 @@ def default_due_date_str(reference_datetime: datetime = None) -> str:
 
 
 def _task_status_from_pull_request(pull_request: PullRequest) -> str:
-    if not pull_request.closed():
+    if not pull_request.closed() and pull_request.is_draft():
+        return "Draft"
+    elif not pull_request.closed() and not pull_request.is_draft():
         return "Open"
     elif pull_request.closed() and pull_request.merged():
         return "Merged"
@@ -100,7 +102,7 @@ _custom_fields_to_extract_map = {
 def _custom_fields_from_pull_request(pull_request: PullRequest) -> Dict:
     """
     We currently expect the project to have two custom fields with its corresponding enum options:
-        • PR Status: "Open", "Closed", "Merged"
+        • PR Status: "Open", "Draft", "Closed", "Merged"
         • Build: "Success", "Failure"
     """
     repository_id = pull_request.repository_id()
--- a/docs/code_reviews.md
+++ b/docs/code_reviews.md
@@ -19,7 +19,7 @@ Field mapping, GitHub to Asana:
 | PR comment                                        | comment in Asana task. Comes from bot, but `@-mentions` comment author in Asana |
 | PR review                                         | comment in Asana task. Comes from bot, but `@-mentions` review author in Asana |
 | PR participants (author + reviewers + commenters) | followers in Asana task                                      |
-| PR status (open/closed/merged)                    | "PR Status" custom field in Asana task
+| PR status (open/draft/closed/merged)              | "PR Status" custom field in Asana task
 | PR build status (pending/success/failure)         | "Build" custom field in Asana task
 
 #### How a GitHub Pull Request looks like
--- a/src/github/graphql/fragments/FullPullRequest.py
+++ b/src/github/graphql/fragments/FullPullRequest.py
@@ -13,6 +13,7 @@ fragment FullPullRequest on PullRequest {
   }
   closed
   merged
+  isDraft
   mergedAt
   mergeable
   url
--- a/src/github/models/pull_request.py
+++ b/src/github/models/pull_request.py
@@ -145,6 +145,9 @@ class PullRequest(object):
     def is_mergeable(self) -> bool:
         return self.mergeable() == MergeableState.MERGEABLE
 
+    def is_draft(self) -> bool:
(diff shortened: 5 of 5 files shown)


lance-rutkin:
@gregeinfrank I still need to walk through testing this with you, but wanted to get the changes and approach up in front of you before proceeding.

scripts/setup_sgtm_tasks_project.py
```diff
@@ -25,6 +25,7 @@ class CustomField:
         name="PR Status",
         enum_options=[
             EnumOption(name="Open", color="green"),
+            EnumOption(name="Draft", color="grey"),
```
gregeinfrank, line 28: where do we have the list of colors that are options? Just want to double check `"grey"` is an official custom field color

gregeinfrank (commented):
Approach looks good! Please add a test case for draft - and if there are no tests for this function I apologize but please add some 🙏

src/asana/helpers.py
```diff
@@ -75,7 +75,9 @@ def default_due_date_str(reference_datetime: datetime = None) -> str:
 
 
 def _task_status_from_pull_request(pull_request: PullRequest) -> str:
-    if not pull_request.closed():
+    if not pull_request.closed() and pull_request.is_draft():
+        return "Draft"
+    elif not pull_request.closed() and not pull_request.is_draft():
```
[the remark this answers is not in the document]
lance-rutkin: Happy w/ the alternative. I was just following the existing pattern. I can also adjust the `closed()` case to handle both as well.

src/asana/helpers.py
```diff
@@ -75,7 +75,9 @@ def default_due_date_str(reference_datetime: datetime = None) -> str:
...
         return "Open"
     elif pull_request.closed() and pull_request.merged():
         return "Merged"
```
lance-rutkin, line 83: i.e.
```py
elif pull_request.closed():
    return "Merged" if pull_request.merged() else "Closed"
```

I think we can actually just make that an `else` and drop the logger error as I believe it is unreachable.

scripts/setup_sgtm_tasks_project.py
gregeinfrank: 🆒

src/asana/helpers.py
gregeinfrank: Cool yeah either way works for me - consistency is great, so either way you pick, can you make sure it's consistent between open/closed logic?

requirements-dev.txt
```diff
@@ -1,3 +1,4 @@
 black==19.10b0
 moto==1.3.10
 mypy==0.790
+click==8.0.4
```
lance-rutkin, line 4: I see you found the same fix!

test/asana/helpers/test_extract_task_fields_from_pull_request.py
```diff
@@ -696,6 +769,133 @@ def test_does_not_rely_on_github_to_return_reviews_sorted_by_submitted_at_timest
         self.assertEqual(True, task_fields["completed"])
 
 
+@patch("src.dynamodb.client.get_asana_id_from_github_node_id", return_value="0")
+class TestExtractsCustomFieldsFromPullRequest(BaseClass):
```
lance-rutkin, line 773: Let me know if these tests are in line with what you're thinking. It involved quite a bit of mocking, so if they don't accomplish exactly what you had in mind, I'm happy to take another crack at them.

scripts/setup_sgtm_tasks_project.py
```diff
@@ -20,11 +20,14 @@ class CustomField:
     enum_options: List[EnumOption]
 
 
+# If adding a new option, ensure that the color you're selecting is part of the list of colors in this [file](https://github.com/Asana/codez/blob/next-master/asana2/asana/web/ui/colors/color_helpers.ts)
```
gregeinfrank, line 23: 💯

gregeinfrank: Actually, is there a public-facing list anywhere in the API documentation? Non-Asana people won't have access to this reference. If not, maybe inline the options that are available today (even though it'll get out of date)

src/asana/helpers.py
```diff
@@ -76,14 +76,9 @@ def default_due_date_str(reference_datetime: datetime = None) -> str:
 
 def _task_status_from_pull_request(pull_request: PullRequest) -> str:
     if not pull_request.closed():
```
gregeinfrank, line 78: I like this format better! One nit: can you switch them around for easier readability? (rather than first condition being "not closed", have the first condition be `if pull_request.closed()`

test/asana/helpers/test_extract_task_fields_from_pull_request.py
gregeinfrank: hm yeah I think I was mostly looking for unit testing `_task_status_from_pull_request`, though I think having _some_ test that checks for this custom field also makes sense - perhaps a bit of both? Have one test that checks that `"pr_status"` custom field gets set at all, then another for unit testing the logic behind `_task_status_from_pull_request`?

gregeinfrank (changes requested):

scripts/setup_sgtm_tasks_project.py
lance-rutkin: From what I can see, it looks like we don't have the list anywhere in the docs.

The color field for [EnumOption](https://developers.asana.com/docs/enum-option) is just typed as a `string`. Maybe this is an opportunity to improve our docs. I'll raise it in #eng-random or something like that.

For now, I'll just include the inline list.

gregeinfrank (approved):

PR #118, opened 2022-04-05
Pin AWS Terraform version

lance-rutkin:
Pins AWS Terraform to earlier version to get around breaking changes described [here](https://github.com/hashicorp/terraform-provider-aws/issues/23125).

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1202081951483499)


lance-rutkin:
Testing to see if CF option is updated

PR #120, opened 2022-09-19
Use data-asana-dynamic=false for Asana links with alt text

gregeinfrank:
See: https://app.asana.com/0/1198486630539940/1202909536189519/f / https://app.asana.com/0/780306770840675/1202908585602823/f

If someone references an Asana task with a vanity url text, Asana will try to evaluate to the object's title unless `data-asana-dynamic` is set to `false`

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1203010140121164)

### DIFF
--- a/src/markdown_parser.py
+++ b/src/markdown_parser.py
@@ -57,9 +57,6 @@ class GithubToAsanaRenderer(mistune.HTMLRenderer):
         # https://developers.asana.com/docs/rich-text
 
         is_asana_vanity_link = link.startswith("https://app.asana.com") and text
-        if text is None:
-            text = link
-
         asana_tags = 'data-asana-dynamic="false" ' if is_asana_vanity_link else ""
         s = "<a " + asana_tags + 'href="' + self._safe_url(link) + '"'
         return s + ">" + (text or link) + "</a>"
--- a/src/markdown_parser.py
+++ b/src/markdown_parser.py
@@ -51,9 +51,21 @@ class GithubToAsanaRenderer(mistune.HTMLRenderer):
 
         return re.sub(URL_REGEX, urlreplace, text)
 
+    def link(self, link, text=None, title=None):
+        is_asana_vanity_link = link.startswith("https://app.asana.com") and text is not None
+        if text is None:
+            text = link
+
+        asana_tags = 'data-asana-dynamic="false" ' if is_asana_vanity_link else ""
+        s = '<a ' + asana_tags + 'href="' + self._safe_url(link) + '"'
+        if title:
+            s += ' title="' + mistune.escape_html(title) + '"'
+        return s + '>' + (text or link) + '</a>'
+
     # Asana's API can't handle img tags
     def image(self, src, alt="", title=None) -> str:
-        return f'<a href="{src}">{alt}</a>'
+        return self.link(src, text=alt, title=title)
+
 
 
 def convert_github_markdown_to_asana_xml(text: str) -> str:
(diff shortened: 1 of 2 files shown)


src/markdown_parser.py
```diff
@@ -51,9 +51,22 @@ def urlreplace(matchobj: Match[str]) -> str:
...
+        is_asana_vanity_link = (
+            link.startswith("https://app.asana.com") and text is not None
+        )
+        if text is None:
+            text = link
+
+        asana_tags = 'data-asana-dynamic="false" ' if is_asana_vanity_link else ""
+        s = "<a " + asana_tags + 'href="' + self._safe_url(link) + '"'
+        if title:
+            s += ' title="' + mistune.escape_html(title) + '"'
```
mmorearty, line 64: API HTML (a strict subset of HTML) does not allow the `title=` attribute. [public docs](https://developers.asana.com/docs/rich-text)

src/markdown_parser.py
```diff
@@ -51,9 +51,22 @@ def urlreplace(matchobj: Match[str]) -> str:
...
+        return s + ">" + (text or link) + "</a>"
+
     # Asana's API can't handle img tags
     def image(self, src, alt="", title=None) -> str:
```
mmorearty, line 68: Since your `link` function expects `None` as a default value, I think you should (if possible) change this to use `alt=None` instead of `alt=""`. The way you have it, if someone calls `image(src)` won't they end up with `<a data-asana-dynamic="false"></a>"`?

mmorearty (changes requested):

src/markdown_parser.py
gregeinfrank: Hm yeah this is odd - for reference, I'm just overriding the mistune functions (https://github.com/lepture/mistune/blob/master/mistune/renderers.py#L137-L146), so the only code calling these functions are the markdown parser itself. I think I'd like to keep the function signature the same, but instead in `def link` assume vanity Asana URL if text is not None OR text is not ""

mmorearty: That sounds good — or just do a truthy/falsy check, `if text`

src/markdown_parser.py
```diff
@@ -51,9 +51,22 @@ def urlreplace(matchobj: Match[str]) -> str:
 
         return re.sub(URL_REGEX, urlreplace, text)
 
+    def link(self, link, text=None, title=None):
+        # the parser may pass in `title`, but Asana's API does not allow the
+        # `title` attribute and is therefore ignored here.
+        # https://developers.asana.com/docs/rich-text
+
+        is_asana_vanity_link = link.startswith("https://app.asana.com") and text
+        if text is None:
```
mmorearty, line 60: ^ same problem still exists on this line

mmorearty: You should probably use `if not text:`

mmorearty (changes requested):

src/markdown_parser.py
```diff
@@ -51,9 +51,22 @@ def urlreplace(matchobj: Match[str]) -> str:
...
+            text = link
+
+        asana_tags = 'data-asana-dynamic="false" ' if is_asana_vanity_link else ""
+        s = "<a " + asana_tags + 'href="' + self._safe_url(link) + '"'
+        return s + ">" + (text or link) + "</a>"
```
gregeinfrank, line 65: I don't think it matters really since text is used like this (`text or link`).  I think the `text = link` line can be removed all together

mmorearty (approved):

PR #124, opened 2022-11-07
Update automerge comment to be more accurate

gregeinfrank:
When automerge after _approval only_ label is selected, the comment was referencing automerge after tests as well. This PR introduces a new comment to more accurately reflect that automerge label, which does not mention tests!

(see: https://app.asana.com/0/780306770840675/1203239731436292/f)

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1203315587152182)


seanwentzel (approved):
This LGTM and improves the status quo. I imagine it's better to consider these two cases entirely separately though, like if a PR has warning A already but it should have warning B, post warning B anyway.

gregeinfrank:
Ah good point! Maybe there's a clean way to do that, I'll ping here for another look if I change the implementation

gregeinfrank:
@seanwentzel  - I updated the logic to check just the automerge comment that is about to be posted

seanwentzel (commented):
lgtm

PR #125, opened 2022-11-11
Split server_side_encryption for s3 bucket into its own resource

gregeinfrank:
Discussion about this [here](https://github.com/hashicorp/terraform-provider-aws/issues/23106) but tldr is that in later versions of the AWS provider, individual configurations on S3 buckets are split out into their own Terraform resources. This was backported to v3.75.2 (which is what we're currently using) to prep for the breaking change.

I already imported into Asana's remote state so this is a no-op

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1203355580854612)

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -343,12 +343,21 @@ terraform {
 
 resource "aws_s3_bucket" "api_key_bucket" {
   bucket = var.api_key_s3_bucket_name
-  server_side_encryption_configuration {
+
+  # See: https://github.com/hashicorp/terraform-provider-aws/issues/23106#issuecomment-1099401600
+  lifecycle {
+    ignore_changes = [
+      server_side_encryption_configuration
+    ]
+  }
+}
+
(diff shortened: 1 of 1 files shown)


gregeinfrank:
@skeggse  OK if I assign this to you as Terraform AOR? Going to merge though since I already `terraform import`ed the new resources into our state

terraform/main.tf
```diff
@@ -343,12 +343,21 @@ terraform {
...
+resource "aws_s3_bucket_server_side_encryption_configuration" "api_key_bucket_server_side_encryption_configuration" {
+    bucket = aws_s3_bucket.api_key_bucket.bucket
     rule {
       apply_server_side_encryption_by_default {
         kms_master_key_id = aws_kms_key.api_encryption_key.arn
         sse_algorithm     = "aws:kms"
       }
     }
-  }
 }
```
skeggse, line 363: `tf fmt` might be good for this file

skeggse (commented):
lgtm

I'd love to see us upgrade to v4+ on the aws provider for consistency with our other Terraform code!

PR #127, opened 2022-12-27
Remove commit status, just use statusCheckRollup

gregeinfrank:
Fixes: https://app.asana.com/0/1149418478823393/1203565858490862/f

This is happening because there are two types of "status". One is a "commit status" (status _classic_), the other is a check run. If we were _only_ relying on the status and falling back to statusCheckRollup, we were incorrectly assuming the PR was in a "success" state when certain types of checks runs (like we use for Testville) would be ignored in this logic.

[This blog post helped me understand the difference, and how we were mis-using status](https://dev.to/gr2m/github-api-how-to-retrieve-the-combined-pull-request-status-from-commit-statuses-check-runs-and-github-action-results-2cen)

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1203618452841244)

### DIFF
--- a/src/github/graphql/fragments/FullPullRequest.py
+++ b/src/github/graphql/fragments/FullPullRequest.py
@@ -68,9 +68,6 @@ fragment FullPullRequest on PullRequest {
   commits(last: 1) {
     nodes {
       commit {
-        status {
-          state
-        }
         statusCheckRollup {
           state
         }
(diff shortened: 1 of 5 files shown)


suzyng83209 (approved):
ah I actually remember reading through documentation on the differences! Just double checking there that the `statusCheckRollup` is what the article is referring to check runs but with a different name?

gregeinfrank:
Yeah - `statusCheckRollup` reports on _both_ commit status *and* check runs (this is where the "rollup" part comes in). See: https://docs.github.com/en/graphql/reference/objects#statuscheckrollup

gregeinfrank:
Test

PR #134, opened 2023-02-01
Update due date to today if PR is handed off to new assignee

gregeinfrank:
When a PR changes hands, it's because of an approval, request changes, or ready for re-review. Because SGTM currently only sets the due date _once_, for any back-and-forth handoff, the due dates can get overdue quickly. This PR introduces logic to set due date to _today_ when a PR changes assignees (unless SGTM thinks a collaborator has manually changed the due date to later intentionally)

See: https://app.asana.com/0/1149418478823393/1203831261518511/f

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1203877928337987)

### DIFF
--- a/src/asana/controller.py
+++ b/src/asana/controller.py
@@ -1,5 +1,4 @@
 from typing import Optional
-from datetime import datetime
 from . import client as asana_client
 from . import helpers as asana_helpers
 from . import logic as asana_logic
@@ -20,7 +19,9 @@ def create_task(repository_id: str) -> Optional[str]:
         return asana_client.create_task(project_id, due_date_str=due_date_str)
 
 
-def update_task(pull_request: PullRequest, task_id: str):
+def update_task(
+    pull_request: PullRequest, task_id: str, force_update_due_today: bool = False
+):
     task_url = asana_helpers.task_url_from_task_id(task_id)
     pr_url = pull_request.url()
     logger.info(f"Updating task {task_url} for pull request {pr_url}")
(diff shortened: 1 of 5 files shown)


gregeinfrank:
@jchurchill - assigning to you as https://app.asana.com/0/77076599077/102854135132070/f AOR to weigh in mostly on the cultural change. If you want to review the code as well, then great!

src/asana/controller.py
```diff
@@ -33,11 +34,28 @@ def update_task(pull_request: PullRequest, task_id: str):
...
+def _new_due_on_or_none(task: dict, update_task_fields: dict) -> Optional[str]:
+    due_on = None
+    default_due_on = asana_helpers.default_due_date_str()
+
+    if task["due_on"] > default_due_on:
+        # don't update due dates that were potentially manually updated past the default
+        return None
+    elif (task.get("assignee") or {}).get("gid") != update_task_fields.get("assignee"):
+        # if the task is switching assignees, update the due date to today
+        return datetime.now().strftime("%Y-%m-%d")
```
jchurchill, line 55: If someone redirects a PR to someone else because they weren't the right person to receive it, it seems like that would update the due date to today, which seems aggressive. What if all this logic instead encoded:
* If assignee changes, update the due date to today.
* except: never update the due date to an earlier time

jchurchill (approved):
Cultural change makes sense to me. Almost anything is better than letting the due date grow stale as long as action is still being taken on the PR.

Since we're here I had a question about the logic, but feel free to resolve it however you see fit.

src/asana/controller.py
gregeinfrank: Yeah I thought about that (which would basically switch the logic above to return None if the _existing_ due date is tomorrow or later, rather than just later than tomorrow) - but I thought it was more common that someone reviews same-day, which then should re-assign to today for next steps.

However, now thinking about it more - I can see a common use case where someone opens a PR assigned to themself, then self-reviews and cleans up a bit, then assigns it to a peer. In that case, we'd want to keep the "next business day" logic.

I think you're right - I can start with a less aggressive approach and see if feedback comes in at all.

jchurchill: If it's easy to do, giving this logic awareness of the PR's author (so it can compare to the new assignee) could also allow it to be more targeted.

gregeinfrank:
@jchurchill  I went with a different approach -- generally, only change the due date to today if the date is stale and PR is changing assignees (covers the case you pointed out), but force to today when the review comes in as an approval or request changes

jchurchill:
Cool, I like it!

PR #135, opened 2023-03-08
Add follow-up review feature

mattbryant-asana:
This makes it possible to specify a list of Github users that can approve PRs but have them not count as reviewed.

This is required to support the Asana-internal `PR Velocity Bot`, but may be a useful feature for other users.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1204140428956027)

### DIFF
--- a/src/config.py
+++ b/src/config.py
@@ -35,5 +35,7 @@ SGTM_FEATURE__AUTOMERGE_ENABLED = is_feature_flag_enabled(
 )
 SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS = {
     int(github_id)
-    for github_id in os.getenv("SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS").split(",")
+    for github_id in os.getenv("SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS", "").split(
+        ","
+    )
 }
--- a/src/config.py
+++ b/src/config.py
@@ -35,8 +35,8 @@ SGTM_FEATURE__AUTOMERGE_ENABLED = is_feature_flag_enabled(
 )
 SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS = {
     github_username
-    for github_username in os.getenv("SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS", "").split(
-        ","
-    )
+    for github_username in os.getenv(
+        "SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS", ""
+    ).split(",")
     if github_username
 }
--- a/src/config.py
+++ b/src/config.py
@@ -34,9 +34,9 @@ SGTM_FEATURE__AUTOMERGE_ENABLED = is_feature_flag_enabled(
     "SGTM_FEATURE__AUTOMERGE_ENABLED"
 )
 SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS = {
-    int(github_id)
-    for github_id in os.getenv("SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS", "").split(
+    github_username
+    for github_username in os.getenv("SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS", "").split(
         ","
     )
-    if github_id
+    if github_username
 }
--- a/src/config.py
+++ b/src/config.py
@@ -38,4 +38,5 @@ SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS = {
     for github_id in os.getenv("SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS", "").split(
         ","
     )
+    if github_id
 }
--- a/src/config.py
+++ b/src/config.py
@@ -33,3 +33,7 @@ SGTM_FEATURE__AUTOCOMPLETE_ENABLED = is_feature_flag_enabled(
 SGTM_FEATURE__AUTOMERGE_ENABLED = is_feature_flag_enabled(
     "SGTM_FEATURE__AUTOMERGE_ENABLED"
 )
+SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS = {
+    int(github_id)
+    for github_id in os.getenv("SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS").split(",")
+}
--- a/src/github/logic.py
+++ b/src/github/logic.py
@@ -1,11 +1,15 @@
 import re
 from typing import List
 from src.logger import logger
+
 from . import client as github_client
-from src.github.models import PullRequest, MergeableState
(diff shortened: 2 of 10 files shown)


mattbryant-asana:
@gregeinfrank I didn't implement any tests yet, since I'd like an initial review on the feature first.

I also didn't change https://github.com/Asana/SGTM/blob/06f3cb4c1628a36519fbf30d8b0760c6fea0809f/src/asana/helpers.py#L367-L384 because I couldn't decide what one would want there.  Seems reasonable to start without that and see if we can clarify things later.

src/github/logic.py
```diff
@@ -69,23 +80,53 @@ def _pull_request_commenters(pull_request: PullRequest) -> List[str]:
...
     The pull request has been approved if the last review (approval/changes
-    requested) before merging was an approval
+    requested) before merging was an approval, ignoring reviews from users that
+    are marked as needing follow-up review.
     """
+    assert pull_request.merged(), "Checked for pre-merge approval on a non-merged PR"
     merged_at = pull_request.merged_at()
-    if merged_at is not None:
-        premerge_reviews = [
+    if merged_at is None:
```
gregeinfrank, line 93: Is this a state we could actually get in, especially with the assertion added above?

gregeinfrank (commented):
Great - this looks like the right approach to me. Adding tests plus manual testing (I can help with that if you haven't figured out how to set up locally) and I think we're good.

src/github/logic.py
mattbryant-asana: Unclear - there are separate `merged` and `merged_at` properties, so it seems possible but unlikely that we could hit this.

mattbryant-asana:
I fixed all the old tests and added a few new tests.  I could definitely use help with manual testing.

test/asana/helpers/test_extract_task_fields_from_pull_request.py
```diff
@@ -255,12 +258,36 @@ def test_html_body_status_closed_approved_before(self, approved_before_merging):
...
+            "BODY",
+            "</body>",
+        ]
+        self.assertContainsStrings(actual, expected_strings)
+
     @patch("src.github.logic.pull_request_approved_before_merging")
     @patch("src.github.logic.pull_request_approved_after_merging")
     def test_html_body_status_closed_approved_after(
-        self, approved_before_merging, approved_after_merging
+        self, approved_after_merging, approved_before_merging
```
mattbryant-asana, line 288: The order of the arguments from `@patch` was incorrect, so this test was checking the wrong thing

gregeinfrank: whoops

test/asana/helpers/test_extract_task_fields_from_pull_request.py
```diff
@@ -424,6 +453,64 @@ def test_completed_is_false_if_pr_is_closed_and_was_approved_before_merging_but_
         )
         self.assertEqual(False, task_fields["completed"])
 
+    @patch(
+        "src.github.logic.SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS", [followup_bot]
```
gregeinfrank, line 457: I'm kind of confused how this is working - since `SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS` should be a set of integers, but `[followup_bot]` is an array of `User`s, how does this code work?

gregeinfrank (commented):
I think this looks good but confused about the patching that I'd like to understand better before approving

test/asana/helpers/test_extract_task_fields_from_pull_request.py
mattbryant-asana: You know, that's a great question...

gregeinfrank: @mattbryant-asana  check out my latest commit -- the one test was using `merged` rather than `merged_at` (I think hitting into the new condition you added where there's no `merged_at` value)

the second was passing because it didn't really matter what the value of the followup user was since it's more of a regression test to make sure no functionality is introduced to prevent completing the task if a human already reviewed before the followup user

gregeinfrank (approved):

test/asana/helpers/test_extract_task_fields_from_pull_request.py
```diff
@@ -454,14 +454,14 @@ def test_completed_is_false_if_pr_is_closed_and_was_approved_before_merging_but_
...
     @patch(
-        "src.github.logic.SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS", [followup_bot]
+        "src.github.logic.SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS", [followup_bot.id()]
     )
     def test_completed_is_false_if_pr_is_closed_and_was_approved_by_followup_user(self):
         pull_request = build(
             builder.pull_request()
             .closed(True)
             .merged(True)
-            .merged("2020-01-13T14:59:59Z")
```
gregeinfrank: I'm surprised that mypy didn't fail here... Since `merged` expects a `bool`, I would expect that to have been an error - I need to check up on that, since I do run mypy during CircleCI builds, which would have helped catch this

mattbryant-asana: `mypy` is quite happy here - I ran it on the `src/` and `test/` directories.

gregeinfrank: Yeah, same - but it doesn't make sense to me - I know I'm on an old version of mypy but even on the newest version that error doesn't show up (but a bunch of other warnings do)

src/config.py
```diff
@@ -33,3 +33,10 @@ def is_feature_flag_enabled(flag_name: str) -> bool:
 SGTM_FEATURE__AUTOMERGE_ENABLED = is_feature_flag_enabled(
     "SGTM_FEATURE__AUTOMERGE_ENABLED"
 )
+SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS = {
+    int(github_id)
```
mattbryant-asana, line 37: Looks like we don't load `id`s for authors (though we mock them out).  I'm going to switch this to take `user.login` instead - things will be a bit less unique since it's effectively a username which can collide for Github Apps, but I can't see that affecting people and it's not really a problem if it does collide.

mattbryant-asana:
Updated to use `user.login()` and tested that it doesn't reassign open tasks that were approved by follow-up users.

src/config.py
gregeinfrank: Ah - if you wanted to use ids you'd have to just add `id` to the author fragments like https://github.com/Asana/SGTM/blob/1d1c308f9754526c08ef86d8ce66958ecbc720b1/src/github/graphql/fragments/FullReview.py#L9

gregeinfrank:
LGTM

src/config.py
mattbryant-asana: I considered that, but decided that I'd have to add it to all the fragments to feel confident, and `login` seems good enough + is already enforced to exist.

PR #136, opened 2023-03-11
Upgrade mypy and fix errors

gregeinfrank:
In #135  we noticed that typechecking didn't work as we'd expected for `merged` vs `merged_at`. This PR upgrades mypy (which _still_ doesn't fix the issue but at least gets us to the latest and greatest so I can take a stab at fixing the real issue)

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1204159273198845)

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,4 +1,4 @@
 black==22.3.0
 moto==1.3.10
-mypy==0.940
+mypy==1.1.1
 types-mock==4.0.12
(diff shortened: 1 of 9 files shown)


gregeinfrank:
In #135  we noticed that typechecking didn't work as we'd expected for `merged` vs `merged_at`. This PR upgrades mypy (which _still_ doesn't fix the issue but at least gets us to the latest and greatest so I can take a stab at fixing the real issue)

gregeinfrank:
By the way, I figured out why those `mypy` errors weren't being caught. By default, mypy will ignore any untyped function defs (like the tests themselves). If you run with `mypy --check-untyped-defs` then it would have caught it. I'll  push out a PR with type fixes at some point when I have time.

mattbryant-asana (approved):

PR #138, opened 2023-03-13
Fix follow-up reviews not being closed after `LGTM` comment

mattbryant-asana:
We short-circuited too soon and never looked for whether the PR had been post-merge approved after being approved by a follow-up user.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1204171313719534)


src/asana/helpers.py
gregeinfrank, line 1: Test comment on whole file

PR #140, opened 2023-04-11
Handle relative URLs

gregeinfrank:
Asana API doesn't handle <a> hrefs that point to a URL without a scheme, so any "link" without a scheme/protocol such as:

* asana.com
* www.asana.com
* /reference/to/relative/file.md

Will be ignored and rendered as text

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1204374742220051)


ericwalton-asana (approved):
LGTM

PR #143, opened 2023-05-08
Add followers incrementally, allowing people to remove themselves from an SGTM task

gregeinfrank:
Currently, any time an SGTM tasks gets updated, we look for _all_ participants on the PR (every comment/review etc) to add followers. This means that someone who comments on a PR can _never remove themself from the SGTM task_ since any new updates would re-add them.

This PR introduces a more incremental / selective approach to adding followers. It only adds comment followers for a comment when a comment is created/updated  for example, rather than looking at _every_ comment every time. This should also help with the efficiency of the function a bit.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1204553465145631)

### DIFF
--- a/src/github/logic.py
+++ b/src/github/logic.py
@@ -226,25 +226,6 @@ def pull_request_participants(pull_request: PullRequest) -> List[str]:
     )
 
 
-def all_pull_request_participants(pull_request: PullRequest) -> List[str]:
-    return list(
-        set(
-            gh_handle
-            for gh_handle in (
-                [pull_request.author_handle()]
-                + pull_request.assignees()
-                + pull_request.reviewers()
-                + pull_request.requested_reviewers()
-                + _pull_request_commenters(pull_request)
-                + _pull_request_comment_mentions(pull_request)
-                + _pull_request_review_mentions(pull_request)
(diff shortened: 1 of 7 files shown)


src/github/logic.py
```diff
@@ -193,18 +211,14 @@ def pull_request_approved_after_merging(pull_request: PullRequest) -> bool:
...
         set(
             gh_handle
             for gh_handle in (
                 [pull_request.author_handle()]
                 + pull_request.assignees()
-                + pull_request.reviewers()
                 + pull_request.requested_reviewers()
-                + _pull_request_commenters(pull_request)
-                + _pull_request_comment_mentions(pull_request)
-                + _pull_request_review_mentions(pull_request)
```
suzyng83209, line 207: from what I can tell, these helper functions aren't used anymore so I think it would be best to delete them in this PR

suzyng83209 (approved):
nit on cleaning up unused functions but otherwise lgtm

gregeinfrank:
test without mentioning Luis

gregeinfrank:
Testing @luis-e-alvarez

gregeinfrank:
Test again

PR #148, opened 2023-06-05
Revert "Rerun required check runs when they exceed a staleness duration"

suzyng83209:
Reverts Asana/SGTM#145

`FullPullRequest` fragment was failing on other queries because it uses `$id` while GetPullRequestAndReview/GetPullRequestAndComment/etc. use `$pullRequestId` variable.

Another issue is that one of the GetPullRequestAnd... queries doesn't even pass in a pull request Id, just commit so I might need to restructure the queries and retry my PR.

Not urgent because I already re-deployed before this changeset.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1204763390125253)


gregeinfrank (approved):
LGTM, Curious why this didn't fail during manual testing though

suzyng83209:
@gregeinfrank I definitely shoe-horned on just the `get_pull_request` function. I wonder if we should have a test for graphql queries against real fake data.

PR #149, opened 2023-06-06
don't add followers if there are none

suzyng83209:
I noticed a few errors in the logs when `update_task` was being invoked. The asana client returns a `ValueError` when there is either an empty array of followers or an array of empty strings. I'll cover the first case and see if I need to cover the second.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1204763785273079)

### DIFF
--- a/src/asana/controller.py
+++ b/src/asana/controller.py
@@ -41,7 +41,8 @@ def update_task(
     if new_due_on is not None:
         update_task_fields["due_on"] = new_due_on
     asana_client.update_task(task_id, update_task_fields)
-    asana_client.add_followers(task_id, followers)
+    if len(followers) > 0:
+        asana_client.add_followers(task_id, followers)
     maybe_complete_tasks_on_merge(pull_request)
 
 


gregeinfrank:
I'm curious how we can get into this situation... There should always be at least one follower added (since we add all mentioned users _and_ the author or commenter themself). Do you have an idea of what would trigger this? Perhaps a comment from someone who isn't in the Github -> asana user id mapping?

suzyng83209:
@gregeinfrank If I'm reading the [logs](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fsgtm/log-events/2023$252F06$252F05$252F$255B$2524LATEST$255Df95a7b41ebaf4b79b9bf36a967f15400$3Fstart$3D1685955432909$26refEventId$3D37598062527612074411172642852157883018512690978640101393) correctly, it looks like this pr https://github.com/Asana/asanadotcom/pull/7565 is one of the prs that caused an error.

the reviewer has an entry in the `sgtm-users` table but there are no other participants in the PR (other than bots). I don't have asanadotcom admin permissions but I suspect that there are no followers being passed in otherwise we would be receiving a different error message

gregeinfrank (approved):
That makes sense - those PRs are opened by bots so the author wouldn't have a mapping

PR #150, opened 2023-06-06
add optional parameters to all queries

suzyng83209:
This is 1 approach to fixing the query issue with missing parameters. I'm ambivalent about this approach because it adds a lot of confusing logic to the query fragments.

I'm choosing between this approach and another approach where I split and unroll `FullPullRequest` and compose the `FullCommit` fragment on the queries that have the parameters it needs. That approach feels more intuitive although it will increase the amount of code duplication since the `FullPullRequest` fragment will need to be broken up.

To my knowledge, there's no way to make a "Higher Order Fragment". Not an expert in graphql so I'm willing to lean either way.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1204771347839821)


src/github/webhook.py
```diff
@@ -130,6 +130,7 @@ def _handle_status_webhook(payload: dict) -> HttpResponse:
         return HttpResponse("200")
 
     with dynamodb_lock(pull_request.id()):
+        pull_request = graphql_client.get_pull_request(pull_request.id())
```
gregeinfrank, line 133: Why pull the pull_request a second time here?

gregeinfrank (commented):
I'm not really sure I follow what this PR is doing and what problem it solves - can you describe more the "query issue with missing parameters"? I'm also definitely not a Graphql expert, so I don't think I have much to add on that front unforunately :(

src/github/webhook.py
suzyng83209: There's no way for me to get the `CheckRun.isRequired` field without either a PR number of a PR id neither of which is available in the webhook payload. Here, I'm remaking the request to get a fully formed PR object with checkrun information.

suzyng83209 (commented):
We have 5 queries to get a `FullPullRequest`:
1. `GetPullRequest` uses `$id` which I pass into the `CheckRun.isRequired(pullRequestId: $id)` query
2. `GetPullRequestAndComment` which uses `$pullRequestId`.
3. `GetPullRequestAndReview` which uses `$pullRequestId`
4. `GetPullRequestByRepositoryAndNumber` which uses $number
5. `GetPullRequestForCommit` which uses a commit id `$id`

Changes:
1. The original issue was the the `FullPullRequest` fragment was failing on 2-5 so I changed the variable to be $pullRequestId
2. I let `isRequired` take `pullRequestNumber` should that be available. However, we have to specify all variables at the top of the query so all queries that use the `FullPullRequest` fragment need both `$pullRequestNumber` and `$pullRequestId` as parameters.
3. Since `GetPullRequestForCommit` doesn't have either variable available, I added a third conditional parameter `$getCheckSuites` to use a directive to conditionally query for checkSuites.
4. Separated out the commit query into its own fragment and cleaned up some extraneous fields/models

PR #151, opened 2023-06-08
Check Run Rerequest logic try #2

suzyng83209:
## Summary
1. [old] Query changes to FullPullRequest to pull down checkSuites and checkRuns(completedAt, databaseId) fields
   - [new] baseRefName field on `PullRequest` object
2. [old] new models for CheckRun and CheckSuite (and their test builders)
3. [old] new logic for maybe merge pull request and refactoring it
4. [new] feature flag `SGTM_FEATURE__CHECK_RERUN_BASE_REF_NAMES` is a list of base ref names to restrict the check run rerequests of
   - this is different from #145 because we removed the `isRequired` field and are instead rerequesting all check runs of a pull request with a base branch of `baseRefName` instead of looking for required checks. `isRequired` query would have resulted in a large graphql query rewrite which is out of scope.

## Test Plan
- [x] new test cases pass 
- [x] manually ran `_maybe_rerun_checks` on test PR https://github.com/Asana/codez/pull/174894
- [x] manually ran `maybe_automerge` on test PR https://github.com/Asana/codez/pull/174894 (expecting failure)
- [x] manually ran all graphql client queries (`GetPullRequest`, `GetPullRequestAndComment`, etc)

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1204783463596101)


terraform/variables.tf
```diff
@@ -71,3 +71,15 @@ variable "sgtm_feature__followup_review_github_users" {
...
+variable "sgtm_feature__check_rerun_threshold_hours" {
+  type = number
+  description = "Number of hours after which a check run should be rerequested. Must be combined with sgtm_feature__check_rerun_base_ref_names."
+  default = 0
+}
+
+variable "sgtm_feature__check_rerun_base_ref_names" {
+  type        = string
+  description = "A comma-separated list of base refs that will trigger check run rerequests when stale. Must be combined with sgtm_feature__check_rerun_freshness_duration_hours."
+  default     = ""
```
gregeinfrank, line 84: Do you think the default here should be `"main,master"`?

gregeinfrank (approved):

terraform/variables.tf
suzyng83209: I think I'll leave this disabled by default. Rerunning check runs is probably not expected behaviour for most workflows (and may incur unexpected costs)

gregeinfrank: I was thinking more that most orgs only open PRs against `main` or `master`, so defaulting that without defaulting the staleness window still doesn't enable the feature, but if you set a `SGTM_FEATURE__CHECK_RERUN_THRESHOLD_HOURS`, that would enable the feature without having to _also_ set a branch name unless you had a unique flow (like ours). Just a thought though, either way works!

PR #154, opened 2023-08-17
Swallow asana api errors

suzyng83209:
When certain task fields can't be updated, we want to try to update the remainder of the fields and be more robust against api down time/api errors.

This shouldn't negatively impact overall performance or error reporting because SGTM frequently doesn't throw errors on lambda and instead returns webhook responses which more often than not time out.

I'm not sure how permanent I want this "solution" to be. But it should give a better user experience for SGTM users in the short term.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1205291532485373)

### DIFF
--- a/src/asana/client.py
+++ b/src/asana/client.py
@@ -2,6 +2,7 @@ from typing import List, Iterator, Dict, Optional
 from typing_extensions import Literal
 import asana  # type: ignore
 from src.config import ASANA_API_KEY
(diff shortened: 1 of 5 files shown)


vn6 (approved):
lgtm! are we planning to merge in these changes ?

suzyng83209:
closing now that we have #155

PR #155, opened 2023-08-29
Add href to asana anchor workaround

suzyng83209:
# Add href to asana anchor workaround

## Summary
The asana client cannot properly handle links to deleted objects that directly reference the global id of the object. This manifests in errors that are difficult to handle when PR bodies contain references to de-activated users. Most frequently, this results is a slew of SGTM errors after intern season.

Instead, we can create an actual link to the github handle with a `data-asana-gid` attribute. If the user_id is valid, the anchor tag should not _do_ anything and be replaced with the appropriate asana task link. If the user_id is _not_ valid, then we should be able to use the anchor tag's href as a valid link.

## Test plan
- [x] a PR that references a de-activated asana user will properly link the github handle link and succeed in upserting the task
- [x] a PR that references an activated user still references the asana task and not the github handle

image
https://app.asana.com/0/1127984365697766/1205259139814521

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1205380760704777)

### DIFF
--- a/src/asana/helpers.py
+++ b/src/asana/helpers.py
@@ -180,7 +180,7 @@ def _asana_user_url_from_github_user_handle(github_handle: str) -> Optional[str]
     user_id = _asana_user_id_from_github_handle(github_handle)
     if user_id is None:
         return None
-    return f'<a data-asana-gid="{user_id}"/>'
+    return _wrap_in_tag("A", attrs={"href": f"https://github.com/{github_handle}", "data-asana-gid": user_id})(github_handle)
 
 
 def _task_name_from_pull_request(pull_request: PullRequest) -> str:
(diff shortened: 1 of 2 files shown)


src/asana/helpers.py
```diff
@@ -180,7 +180,13 @@ def _asana_user_url_from_github_user_handle(github_handle: str) -> Optional[str]
     user_id = _asana_user_id_from_github_handle(github_handle)
```
vn6, line 180: might be helpful to add description for this function like there is `_asana_display_name_for_github_user`

vn6 (approved):
lgtm

src/asana/helpers.py
suzyng83209: oh this is actually getting the asana `gid` from the dynamodb table not the display name which itself is a wrapper around `dynamodb_client.get_asana_domain_user_id_from_github_handle(github_handle)` but I'll remove the wrapper actually since it doesn't serve a purpose

PR #158, opened 2023-11-06
Add Review Status To SGTM Sync

drewalam-asana:
- added review status CF to SGTM sync. There are 4 possible values: Changes Requested, Not Ready, Approved, Needs Review. This will help folks figure out if a PR needs to be reviewed or not.
- modified the approved status for Pull Request objects to utilize the latest review's status to determine if a PR is considered to be "Approved" or "Changes Requested"

### Testing
General format
```
In [1]: import src.github.controller as github_controller; import src.github.graphql.client as graphql_client

In [2]: pull_request = graphql_client.get_pull_request("PR_kwDOAr0gsc5elLoP")

In [3]: github_controller.upsert_pull_request(pull_request)
```

I tested the controller with a [PR ](https://github.com/Asana/codez/pull/224261)I already had up that had the changes requested status. I was able to modify the Asana task to reflect this new status (changes request).
I then asked someone else to approve the pr changes to approved.
I then asked someone else to request changes and the pr status changes and it goes back to needs changes. Task logs:
image

I created a draft PR too and the Draft status was also set. Also tested going from needs review -> draft on the pr and works as expected.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1205891141351354)

### DIFF
--- a/scripts/setup_sgtm_tasks_project.py
+++ b/scripts/setup_sgtm_tasks_project.py
@@ -80,6 +80,7 @@ CUSTOM_FIELDS = [
             EnumOption(name="Needs Review", color="yellow"),
             EnumOption(name="Changes Requested", color="red"),
             EnumOption(name="Approved", color="green"),
+            EnumOption(name="Draft", color="blue")
         ],
     ),
 ]
--- a/scripts/setup_sgtm_tasks_project.py
+++ b/scripts/setup_sgtm_tasks_project.py
@@ -80,7 +80,7 @@ CUSTOM_FIELDS = [
             EnumOption(name="Needs Review", color="yellow"),
             EnumOption(name="Changes Requested", color="red"),
             EnumOption(name="Approved", color="green"),
-            EnumOption(name="Draft", color="blue"),
+            EnumOption(name="Draft", color="purple"),
         ],
     ),
 ]
--- a/scripts/setup_sgtm_tasks_project.py
+++ b/scripts/setup_sgtm_tasks_project.py
@@ -78,7 +78,7 @@ CUSTOM_FIELDS = [
         name="Review Status",
         enum_options=[
             EnumOption(name="Needs Review", color="yellow"),
-            EnumOption(name="Requested Changes", color="red"),
+            EnumOption(name="Changes Requested", color="red"),
             EnumOption(name="Approved", color="green"),
         ],
     ),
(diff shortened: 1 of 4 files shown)


src/asana/helpers.py
```diff
@@ -84,6 +84,26 @@ def _task_status_from_pull_request(pull_request: PullRequest) -> str:
...
+        ),
+        key=lambda r: r.submitted_at(),
+    )
+
+    if len(approval_or_changes_requested_reviews) == 0:
+        # We didn't find any reviews
+        return "Needs Review"
+
+    latest_review = approval_or_changes_requested_reviews[-1]
+    if latest_review.is_approval():
```
suzyng83209, line 102: Does this account for the state where one person requested changes and another person approves later?

SGTM currently waits for all reviewers that requested changes to approve before it determines that the PR has been "approved" instead of looking at the last review. You can checkout `pull_request.is_approved` logic

drewalam-asana: I don't think this makes sense to wait for everyone to approve the PR to mark the PR as approved. 
I think it makes more sense to refer to the latest review left on the PR. What if user A requests changes that are not actually needed and user B follows up saying it's okay and approves the changes. If user B does agree with user A's comments, they'll just add in their additional comments or also request changes as well.

Thoughts?

drewalam-asana:
Current dependencies on/for this PR:
* master
  * **PR #158** <a href="https://app.graphite.dev/github/pr/Asana/SGTM/158?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Asana/SGTM/158?utm_source=stack-comment).

scripts/setup_sgtm_tasks_project.py
```diff
@@ -74,6 +74,15 @@ def resource_subtype(self):
         ],
     ),
     PeopleCustomField(name="Author (SGTM)"),
+    EnumCustomField(
+        name="Review Status",
+        enum_options=[
+            EnumOption(name="Needs Review", color="yellow"),
+            EnumOption(name="Changes Requested", color="red"),
+            EnumOption(name="Approved", color="green"),
+            EnumOption(name="Draft", color="blue")
```
suzyng83209: 🤔 we already have a `Draft` enum for `PR Status`. Can this just be unpopulated if its a draft?
image

suzyng83209 (changes requested):

suzyng83209 (approved):
LGTM! Thanks for doing all the user research and following up with the different avenues I suggested. 

One thing to double check here is error handling. Will the task update/creation fail if the project doesn't have the custom field? If so, we probably need to update all SGTM projects to have this custom field before we deploy this change

drewalam-asana:
When I accidentally tried to test this PR on an Asana tasks for the SGTM repo, it didn't populate correctly so it's safe. I'll go in and modify every SGTM project to have this field though.

PR #160, opened 2024-02-01
Add logging and move graphql queries out of lock

michael-huang87:
Logs the status of PRs to explain why PR is not ready for automerge
Logs when function to update asana task is called
Logs PR node id and comment url when issue_comment webhook is called

Moves graphql queries out of dynamodb lock where possible

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1206499382584755)

### DIFF
--- a/src/github/controller.py
+++ b/src/github/controller.py
@@ -11,6 +11,7 @@ from src.config import (
 
 
 def upsert_pull_request(pull_request: PullRequest):
+    logger.info("Updating/creating asana task for pull request")
     pull_request_id = pull_request.id()
     task_id = dynamodb_client.get_asana_id_from_github_node_id(pull_request_id)
     if task_id is None:
--- a/src/github/controller.py
+++ b/src/github/controller.py
@@ -11,7 +11,6 @@ from src.config import (
 
 
 def upsert_pull_request(pull_request: PullRequest):
-    logger.info("Updating/creating asana task for pull request")
     pull_request_id = pull_request.id()
     task_id = dynamodb_client.get_asana_id_from_github_node_id(pull_request_id)
     if task_id is None:
--- a/src/github/logic.py
+++ b/src/github/logic.py
@@ -288,7 +288,9 @@ def maybe_automerge_pull_request_and_rerun_stale_checks(
             pull_request.is_mergeable() and pull_request.is_approved()
         )
 
-    logger.info(f"Pull request status: test {pull_request.is_build_successful()}, mergeable {pull_request.is_mergeable()}, approved {pull_request.is_approved()}")
+    logger.info(
+        f"Pull request status: test {pull_request.is_build_successful()}, mergeable {pull_request.is_mergeable()}, approved {pull_request.is_approved()}"
+    )
     logger.info(
         f"{pull_request.id()} is {'' if is_pull_request_ready_for_automerge else 'not '}ready for automerge"
     )
--- a/src/github/logic.py
+++ b/src/github/logic.py
@@ -215,6 +215,7 @@ def pull_request_participants(pull_request: PullRequest) -> List[str]:
 def maybe_add_automerge_warning_comment(pull_request: PullRequest):
     """Adds comment warnings if automerge label is enabled"""
 
+    logger.info("Running maybe add automerge warning comment")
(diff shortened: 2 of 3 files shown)


src/github/controller.py
```diff
@@ -11,6 +11,7 @@
 
 
 def upsert_pull_request(pull_request: PullRequest):
+    logger.info("Updating/creating asana task for pull request")
```
suzyng83209, line 14: If this useful considering we have logs for the if...else case on lines 23 and 28?

src/github/logic.py
```diff
@@ -288,6 +288,9 @@ def maybe_automerge_pull_request_and_rerun_stale_checks(
             pull_request.is_mergeable() and pull_request.is_approved()
         )
 
+    logger.info(
+        f"Pull request status: test {pull_request.is_build_successful()}, mergeable {pull_request.is_mergeable()}, approved {pull_request.is_approved()}"
```
suzyng83209, line 292: Can you prepend this with the id too? Also `test` is probably difficult to understand out of context considering the value it likely to be `true` or `false`. These logs should also be well-formed to serve other companies (whether they use SGTM or not) so it would be better to have more explicit logs

src/github/webhook.py
```diff
@@ -89,31 +89,32 @@ def _handle_pull_request_review_comment(payload: dict):
...
+    elif action == "deleted":
+        pull_request = graphql_client.get_pull_request(pull_request_id)
+        review = graphql_client.get_review_for_database_id(
+            pull_request_id, review_database_id
+        )
+    else:
+        raise ValueError(f"Unexpected action: {action}")
 
+    with dynamodb_lock(pull_request_id):
+        if review is None:
```
suzyng83209, line 111: Does this need to be a separate `if` statement if we have line 116? `if...else` should cover that right?

src/github/webhook.py
```diff
@@ -89,31 +89,32 @@ def _handle_pull_request_review_comment(payload: dict):
...
+            # If we deleted the last comment from a review, Github might have deleted the review.
+            # If so, we should delete the Asana comment.
+            logger.info("No review found in Github. Deleting the Asana comment.")
+            github_controller.delete_comment(comment_id)
         if review is not None:
+            logger.info("Updating review comment")
```
suzyng83209, line 117: What's the thinking behind the log statements here vs. inside the `github_controller` methods? Or the client methods?

src/github/webhook.py
```diff
@@ -14,8 +14,8 @@
 # https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request
 def _handle_pull_request_webhook(payload: dict) -> HttpResponse:
     pull_request_id = payload["pull_request"]["node_id"]
+    pull_request = graphql_client.get_pull_request(pull_request_id)
```
suzyng83209, line 17: Why are we not logging pull_request here when we do it for issue_comments on line 29?

suzyng83209 (changes requested):

src/github/webhook.py
michael-huang87: I added this log to match with the part above, but it seems like there is already logging in the github_controller. Its not consistent, `githubcontroller.upsert_review` has logging but `github_controller.delete_comment` does not

src/github/webhook.py
michael-huang87: the pull_request info is logged when we call `maybe_automerge_pull_request_and_rerun_stale_checks` but the comment info is never logged in the code paths it takes

suzyng83209: We're only logging 3 pull request fields. Can we log the entire json body? I suspect we'll want to view it since the granularity of the log you added is useful for clear-cut cases but we may want to know which status checks failed which caused the automerge to fail

suzyng83209 (changes requested):
just one change around more granular logging.

src/github/webhook.py
michael-huang87: Added logging of pull_request json to the logic function as opposed to logging in each of the webhooks

suzyng83209 (approved):
Nice. Lets pair on deploying the new lambda. Drop some time on my cal today?

PR #161, opened 2024-02-01
Autorerun stale checks on approval

vn6:
- Add feature under flag to rerun stale checks if PR is approved
- Add unit tests for new functionality

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1206501097102591)


src/github/logic.py
```diff
@@ -19,6 +19,7 @@
     SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS,
     SGTM_FEATURE__CHECK_RERUN_THRESHOLD_HOURS,
     SGTM_FEATURE__CHECK_RERUN_BASE_REF_NAMES,
+    SGTM_FEATURE__ALLOW_CHECK_RERUN_ON_APPROVAL,
```
suzyng83209, line 22: nit: naming convention to more closely match the previous 2 check rerun fields

src/github/logic.py
```diff
@@ -307,11 +304,36 @@ def maybe_automerge_pull_request_and_rerun_stale_checks(
...
+    )
+    return False
+
+
 # ----------------------------------------------------------------------------------
 # Automerge helpers
 # ----------------------------------------------------------------------------------
 
 
+def _pull_request_is_open(pull_request: PullRequest) -> bool:
```
suzyng83209, line 333: not a requested change: What's the decision behind adding a helper to determine pull request open state vs. adding a class method to `PullRequest` to do it and to query it as a field? i.e. `pull_request.open()` or `pull_request.is_open()`

src/github/logic.py
```diff
@@ -307,11 +304,36 @@ def maybe_automerge_pull_request_and_rerun_stale_checks(
...
+        SGTM_FEATURE__ALLOW_CHECK_RERUN_ON_APPROVAL
+        and _pull_request_is_open(pull_request)
+        and pull_request.is_approved()
+    ):
+        logger.info(
+            f"PR-{pull_request.id()} is open and approved, maybe rerun stale checks"
+        )
+        return _maybe_rerun_stale_checks(pull_request)
+    logger.info(
+        f"SGTM_FEATURE__ALLOW_CHECK_RERUN_ON_APPROVAL is {SGTM_FEATURE__ALLOW_CHECK_RERUN_ON_APPROVAL}"
```
suzyng83209, line 320: I don't think we log any of the feature flags. The exception being https://github.com/Asana/SGTM/blob/5f83954ace9ab36ae8d336053bfac45ea51c1a38/src/github/logic.py#L326 but this log exists mainly to check the base ref name of the PR. What value does this log line give?

test/github/test_logic.py
```diff
@@ -634,6 +634,93 @@ def test_maybe_rerun_stale_checks_multiple_base_refs(
...
+                .submitted_at("2020-01-13T14:59:58Z")
+                .state(ReviewState.APPROVED)
+            )
+            .merged(False)
+        )
+        self.assertFalse(
+            github_logic.maybe_rerun_stale_checks_on_approved_pull_request(pull_request)
+        )
+        mock_rerequest_check_run.assert_not_called()
+
```
suzyng83209, line 723: Can you add another test case for the intersection of logic between this feature and someone that has the `merge after ...` labels? I think this logic is straight forward but its possible that there are race conditions when these 2 rerun features intersect.

suzyng83209 (changes requested):
nits but biggest thing is adding test cases to see behaviour _with_ automerge labels

src/github/logic.py
vn6: Mainly because the `PullRequest` class methods get values directly from the API, which is not what we're doing here

src/github/logic.py
vn6: was using this for testing but can remove!

test/github/test_logic.py
```diff
@@ -698,7 +698,55 @@ def test_no_rerun_stale_checks_on_unapproved_pull_request(
...
+                builder.label().name(
+                    github_logic.AutomergeLabel.AFTER_TESTS_AND_APPROVAL.value
+                )
+            )
+            .merged(False)
+        )
+        self.assertTrue(
+            github_logic.maybe_rerun_stale_checks_on_approved_pull_request(pull_request)
+        )
+        mock_rerequest_check_run.assert_called_once_with(
```
suzyng83209, line 736: I'm a little surprised that we only expect to rerequest the check_run once. My understanding of the logic is:
1. a PR with a `merge after tests and approval` with stale tests gets approved
2. we re-run it via `rerun after approval`
3. we re-run it again via `autorerun on automerge`

Would you be able to walk through why #3 doesn't get hit?

suzyng83209 (approved):
Looks good but I just want to double check logic in one place

test/github/test_logic.py
vn6: I think 3 would still get hit but only after line 746. At this point, we havent run any automerge logic yet.

suzyng83209: ooh I see. Is this behaviour something we want to retain? We'll be making 2 back-to-back rerun requests to Testville in this pathway which could introduce race conditions

src/github/logic.py
```diff
@@ -277,10 +277,14 @@ def maybe_automerge_pull_request_and_rerun_stale_checks(
...
-            and _maybe_rerun_stale_checks(pull_request)
-        )
+        # if PR has already rerun checks, we don't need to rerun them again
+        if did_rerun_checks:
+            did_rerun_stale_required_checks = did_rerun_checks
+        else:
+            did_rerun_stale_required_checks = (
+                is_pull_request_ready_for_automerge
+                and _maybe_rerun_stale_checks(pull_request)
+            )
```
suzyng83209, line 287: Here's a contentious suggestion:
```suggestion
        if not did_rerun_stale_required_checks:
            did_rerun_stale_required_checks = (
                is_pull_request_ready_for_automerge
                and _maybe_rerun_stale_checks(pull_request)
            )
```
aliasing parameters/variables is not a great pattern because it makes it harder to find these references in the future (for refactoring or just making changes to logic). 
However, overwriting parameters within the scope of the function are also not ideal because it's bad practice in every other language that does pass by reference.

I would prefer to trade the former over the latter but it's up to you

PR #162, opened 2024-02-05
update dynamodb lock and move logic functions out of lock

michael-huang87:
Updates dynamodb lock timeout to 45s, and moves github_logic function calls out of the lock. Only github_controller and asana_controller interact with dynamodb, and github_logic doesn't call either so it does not need to be inside the lock.

Also adding some logging to task update function to try to debug issues we are seeing there

https://app.asana.com/0/1149418478823393/1206470410132273/f

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1206529515761826)

### DIFF
--- a/src/asana/controller.py
+++ b/src/asana/controller.py
@@ -32,6 +32,7 @@ def update_task(
     update_task_fields = asana_helpers.extract_task_fields_from_pull_request(
         pull_request
     )
+    logger.info(f"Updating task {task_id} with fields: {update_task_fields}")
     task = asana_client.get_task(task_id)
     new_due_on = (
         asana_helpers.today_str()
--- a/src/dynamodb/lock.py
+++ b/src/dynamodb/lock.py
@@ -23,7 +23,6 @@ lock_client = DynamoDBLockClient(
 def dynamodb_lock(
     lock_name: str, retry_timeout: Optional[timedelta] = timedelta(seconds=45)
 ):
-    # TODO: Make this match get-lock-client in the clojure code
     lock = lock_client.acquire_lock(
         lock_name, sort_key=lock_name, retry_timeout=retry_timeout
     )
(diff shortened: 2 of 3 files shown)


src/asana/controller.py
```diff
@@ -32,6 +32,7 @@ def update_task(
     update_task_fields = asana_helpers.extract_task_fields_from_pull_request(
         pull_request
     )
+    logger.info(f"Updating task {task_id} with fields: {update_task_fields}")
```
suzyng83209, line 35: this will log in the form of:
```
{'field_name': 'xml/jsonified field body'}
```
fyi. Seems fine to me but it might be a lot of logs

src/dynamodb/lock.py
```diff
@@ -21,7 +21,7 @@
 
 @contextmanager
 def dynamodb_lock(
-    lock_name: str, retry_timeout: Optional[timedelta] = timedelta(seconds=20)
+    lock_name: str, retry_timeout: Optional[timedelta] = timedelta(seconds=45)
 ):
     # TODO: Make this match get-lock-client in the clojure code
```
suzyng83209, line 26: can you delete this TODO? I have no idea what it references

src/github/webhook.py
```diff
@@ -15,10 +15,10 @@
 def _handle_pull_request_webhook(payload: dict) -> HttpResponse:
     pull_request_id = payload["pull_request"]["node_id"]
     pull_request = graphql_client.get_pull_request(pull_request_id)
+    # a label change will trigger this webhook, so it may trigger automerge
+    github_logic.maybe_automerge_pull_request_and_rerun_stale_checks(pull_request)
+    github_logic.maybe_add_automerge_warning_comment(pull_request)
```
suzyng83209, line 20: I actually think we want to keep these components inside the lock. These don't mutate the task but it could mutate the PR.

Same with the other moves

suzyng83209 (changes requested):

suzyng83209 (approved):
I think we may want to update the lambda duration itself soon but we can do it iteratively. LGTM

PR #164, opened 2024-02-23
Allow SGTM to be setup with Terraform Cloud 

suzyng83209:
SGTM has only used the s3-backed terraform backend with terragrunt for a while. Asana is moving to using just Terraform Cloud so we want to enable SGTM to move to TFC as well.

Changes:
* Created a new generate block with conditionals depending on which remote backend wants to use. This should generate a well-formatted `backend.tf`
* Still kept the old pathway for s3 remote backend
* Added 3 new TF vars to define the backend configuration
* Updated docs and instructions for deployment

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1206682923832152)

### DIFF
--- a/terraform/variables.tf
+++ b/terraform/variables.tf
@@ -35,11 +35,13 @@ variable "lambda_runtime" {
 
 variable "terraform_backend_s3_bucket_name" {
   type        = string
+  default     = ""
   description = "S3 bucket name to store the Terraform state"
 }
 
 variable "terraform_backend_dynamodb_lock_table" {
   type        = string
+  default     = ""
   description = "The DynamoDb table to store the Terraform state lock"
 }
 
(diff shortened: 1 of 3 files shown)


terraform/variables.tf
```diff
@@ -43,6 +43,22 @@ variable "terraform_backend_dynamodb_lock_table" {
...
+
+variable "terraform_backend_tfc_organization" {
+  type        = string
+  description = "The Terraform Cloud organization to use as the remote backend. Must be provided if terraform_backend_use_tfc is true."
+}
+
+variable "terraform_backend_tfc_workspace" {
+  type        = string
+  description = "The Terraform Cloud workspace to use as the remote backend. Must be provided if terraform_backend_use_tfc is true."
+}
```
skeggse, line 60: these aren't optional, which might cause some confusion. not sure if terragrunt has a good solution to that, though.

suzyng83209: ah you're right. I'll add defaults with empty strings for now

PR #165, opened 2024-04-12
[Deprecate Asanas in Github project] Attempt to read asana user gids from S3 instead of from DynamoDB

harshita-gupta:
### Summary

We'd like to stop maintaining the [Asanas in Github Asana project](https://app.asana.com/0/64201726605125/64201726605125), which is currently kept up-to-date by a lambda that Infrasec maintains. SGTM is the final remaining dependency on this project. This PR adds code for SGTM to read from S3 to determine which asana user a github username corresponds to. If reading from the S3 file fails for any reason, SGTM will revert to the existing DynamoDB table, which is still being synced from the Asanas in Github project.

I'll be monitoring SGTM logs  [here](https://us-east-1.console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fsgtm/log-events$3FfilterPattern$3D$26start$3D-3600000) to ensure that there's no warnings emitted, and to confirm that all usernames are being converted to gids correctly. Once it's been ~1 business day of things working as expected, I'll make a follow-up PR to remove the backup pathway, and then another one to delete all the unneeded infra :)  

Asana tasks:
https://app.asana.com/0/1149418478823393/1204738744867777/f

Relevant deployment: 

CC: @suzyng83209 @prebeta @vn6 @mattbryant-asana @skeggse

### Risks

Pull Request: https://github.com/Asana/SGTM/pull/165

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1207061300950944)

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,3 +2,5 @@ black==22.3.0
 moto==4.1.7
 mypy==1.1.1
 types-mock==4.0.12
+typeguard>=4
+mypy-boto3-s3~=1.34.0
--- a/src/dynamodb/client.py
+++ b/src/dynamodb/client.py
@@ -1,10 +1,17 @@
+from contextlib import closing
+import json
+import traceback
 from typing import Iterator, Optional, List, Tuple
 from typing_extensions import TypedDict
 
 import boto3  # type: ignore
 from botocore.exceptions import NoRegionError  # type: ignore
 
-from src.config import OBJECTS_TABLE, USERS_TABLE
+from src.config import (
+    OBJECTS_TABLE,
+    USERS_TABLE,
+    GITHUB_USERNAMES_TO_ASANA_GIDS_S3_PATH,
+)
 from src.logger import logger
 from src.utils import memoize
 
@@ -115,10 +122,14 @@ class DynamoDbClient(object):
 
     # USERS TABLE
 
-    def bulk_insert_github_handle_to_asana_user_id_mapping(
+    def DEPRECATED_bulk_insert_github_handle_to_asana_user_id_mapping(
         self, gh_and_asana_ids: List[Tuple[str, str]]
     ):
-        """Insert multiple mappings from github handle to Asana user ids."""
+        """Insert multiple mappings from github handle to Asana user ids.
+
+        This is marked as deprecated since we'd like to read SGTM user info from S3 instead of
+        DynamoDB.
+        """
         items = [
             {
                 self.GITHUB_HANDLE_KEY: {"S": gh_handle},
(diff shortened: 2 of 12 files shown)


harshita-gupta:
see https://github.com/Asana/codez/pull/248943 for the PR where we create the file for SGTM to read from, and where we give SGTM's lambda access to the S3 object via the bucket policy.

requirements-dev.txt
```diff
@@ -2,3 +2,5 @@ black==22.3.0
 moto==4.1.7
 mypy==1.1.1
 types-mock==4.0.12
+typeguard>=4
+mypy-boto3-s3~=1.34.0
```
harshita-gupta, line 6: I updated requirements.txt so that `awsudo` would run happily in my SGTM virtual env

src/dynamodb/client.py
```diff
@@ -176,6 +193,72 @@ def _create_client():
...
+
+    # the singleton instance of S3Client
+    _singleton = None
+
+    def __init__(self):
+        self.s3_client = S3Client._create_s3_client()
+        (
+            self.github_user_mapping_bucket_name,
+            self.github_user_mapping_key_name,
+        ) = GITHUB_USERNAMES_TO_ASANA_GIDS_S3_PATH.split("/", 1)
```
suzyng83209, line 210: providing this path is needed for the lambda to run then right since this assignment will fail unless we get a list of length 2?

Can you update the `main.tf` and `config.py` files to also specify that it is required and to not default to `""`? Or conversely, add error handling here?

suzyng83209 (changes requested):
Everything looks good but I think we could do with a little more error handling/requiring the environment variable exists

harshita-gupta:
* **#166** <a href="https://app.graphite.dev/github/pr/Asana/SGTM/166?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#165** <a href="https://app.graphite.dev/github/pr/Asana/SGTM/165?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @harshita-gupta and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

src/dynamodb/client.py
harshita-gupta: ah good point! yes, I'm just going to have this pathway 'fail' on the env var not being set, so that we revert to 'old pathway' if the env var isn't set. (I went with the anti-pattern of an `except Exception` ) in `get_asana_domain_user_id_from_github_handle` intentionally. 

In #166 I'll make sure that deployment fails if the env var isn't set.

FYI I also updated the instructions here: https://app.asana.com/0/1149418478823393/815237203241118/f to include the new env var.

asana-unito:
**:warning: Reviewer:** If you approve this PR, it will be auto-merged as soon as tests pass. If you don't want this to be auto-merged, either Request Changes or remove the auto-merge label before accepting.

suzyng83209 (approved):
thanks!

PR #169, opened 2024-04-24
Set up staging for SGTM

vn6:
### Summary

Asana tasks:

Relevant deployment: 

CC: @vn6

### Risks

Pull Request: https://github.com/Asana/SGTM/pull/169

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1207155226108365)

### DIFF
--- a/terraform/sgtm-cluster/main.tf
+++ b/terraform/sgtm-cluster/main.tf
@@ -173,7 +173,7 @@ resource "aws_lambda_function" "sgtm" {
 
 resource "null_resource" "install_python_dependencies" {
   triggers = {
-    src_sha1 = sha1(join("", [for f in fileset(path.root, "../../src/**") : filesha1(f)]))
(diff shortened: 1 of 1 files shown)


terraform/sgtm-cluster/main.tf
```diff
@@ -0,0 +1,432 @@
...
+  triggers = {
+    src_sha1 = sha1(join("", [for f in fileset(path.root, "../src/**") : filesha1(f)]))
+  }
+
+  provisioner "local-exec" {
+    command = "bash ${path.module}/../scripts/create_pkg.sh"
+
+    environment = {
+      source_code_path = "../src"
+      function_name    = "sgtm"
```
harshita-gupta, line 184: This might need to be changed to include the suffix, but I'm not 100% sure about that 🤔

vn6: hmm in `scripts/create_pkg.sh` looks like it's only used to set up a virtual env so seems ok to leave this

PR #173, opened 2024-04-29
[SGTM SQS] Add SQS for requeueing failures

vn6:
### Summary
- add sqs to retry failures
- failures will be retried once

Asana tasks:
https://app.asana.com/0/1200740774079340/1206916572348055/f

Relevant deployment: 

CC: @vn6 @harshita-gupta @suzyng83209

### Test Plan
tested with https://github.com/Asana/testing-sgtm/pull/1, see asana task https://app.asana.com/0/1207118214445707/1207185028761070/f
- verify that regular webhook events work
- verify that failure + sqs event works (see screenshot below for logs of this)
image

### Risks

Pull Request: https://github.com/Asana/SGTM/pull/173

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1207199132916045)

### DIFF
--- a/terraform/sgtm-cluster/main.tf
+++ b/terraform/sgtm-cluster/main.tf
@@ -167,7 +167,7 @@ resource "aws_lambda_function" "sgtm" {
 }
 
 locals {
-  dist_dir_name = "lambda_dist_pkg/${local.suffix}"
+  dist_dir_name = "lambda_dist_pkg/pkg${local.suffix}"
 }
 
 resource "null_resource" "install_python_dependencies" {
@@ -192,7 +192,7 @@ resource "null_resource" "install_python_dependencies" {
 data "archive_file" "create_dist_pkg" {
   depends_on  = [null_resource.install_python_dependencies]
   source_dir  = "${path.cwd}/${local.dist_dir_name}"
-  output_path = "build/${local.suffix}/function.zip"
+  output_path = "build/pkg${local.suffix}/function.zip"
   type        = "zip"
 }
 
--- a/src/handler.py
+++ b/src/handler.py
@@ -3,7 +3,6 @@ import hmac
 import json
 import traceback
 import boto3
-
 from src.http import HttpResponse, HttpResponseDict
 from src.config import GITHUB_HMAC_SECRET, SQS_URL
 from src.logger import logger


Pipfile
```diff
@@ -22,6 +22,7 @@ types-mock = "==4.0.12"
 typeguard = ">=4"
 mypy-boto3-s3 = "~=1.34.0"
 mypy-boto3-sso = "*"
+boto3-stubs = "*"
```
vn6, line 25: for passing mypy test

terraform/sgtm-cluster/main.tf
```diff
@@ -407,6 +408,43 @@ resource "aws_lambda_permission" "lambda_permission_for_sgtm_rest_api" {
   source_arn = "${var.sgtm_rest_api_execution_arn}/*/${aws_api_gateway_method.sgtm_post.http_method}${aws_api_gateway_resource.sgtm_resource.path}"
 }
 
+## SQS
+resource "aws_sqs_queue" "sgtm-webhooks-queue-fifo" {
+  name = "sgtm-webhooks-queue${local.suffix}.fifo"
+  fifo_queue = true
+  content_based_deduplication = true
```
suzyng83209, line 415: This is AWS's content dedupe logic right? Should protect against duplicate or similar enough webhooks but were we thinking of doing more advanced deduping? Is there a way to specify a dedupe key/logic?

terraform/sgtm-cluster/main.tf
```diff
@@ -407,6 +408,43 @@ resource "aws_lambda_permission" "lambda_permission_for_sgtm_rest_api" {
...
+  visibility_timeout_seconds = 720  # 12 minutes
```
suzyng83209, line 416: SGTM's lambda timeout is set for 2 minutes right? Wouldn't we want to only lock the message for ~2 minutes with the assumption that if it hasn't been deleted yet then the lambda timed out or something went wrong? 12 minutes is a long time.

suzyng83209 (commented):
I noticed this approach is using the lambda as the entrypoint for the api gateway and follow up sqs events. Just wanted to know if that's what we've settled on because of the technical difficulties with the gateway passthrough and if we're still interested in following up with that approach.

Also, some questions.

terraform/sgtm-cluster/main.tf
vn6: Yep, messages are deduped by the `MessageGroupId` which is currently being set to the webhook `delivery_id`. We were thinking of deduping by PR node id, but would need custom logic to avoid losing different webhook event types. I believe aws does not allow custom deduping logic.

terraform/sgtm-cluster/main.tf
vn6: ah I was following a recommendation from https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#events-sqs-queueconfig
image

The recommendation is for using the SQS built-in retry though, which we aren't doing here so seems reasonable to shorten it. We could consider using the SQS retry if we want to enable multiple retries for each webhook event.

vn6:
> Just wanted to know if that's what we've settled on because of the technical difficulties with the gateway passthrough and if we're still interested in following up with that approach.

yep, not sure it's possible to preserve the webhook body contents when passing from the api gateway to sqs. I can add a more detailed description of the blockers / attempts to fix it in https://app.asana.com/0/1200740774079340/1206916572348055/f.

terraform/sgtm-cluster/main.tf
suzyng83209: SGTM is in a weird spot right now where much of the event handler logic is the same except for a few extra steps or queries or datamodel changes. All events that use the same pull request node id end up waiting for the same lock because these handlers have slightly divergent but shared mutations. 

I think it would be interesting to see if we could isolate the event handler's logic more and be able to lock on different node ids (i.e. updating a PR comment does not lock the PR node id -- although not sure if asana would take that well). Something to look into when we have the bandwidth 😅

terraform/sgtm-cluster/main.tf
suzyng83209: hmm yeah I'm not sure how applicable this is to us. The main concern around decreasing the visibility timeout is increasing the number of lockable events for dynamodb and increasing the rate of lock timeouts whereas the concern for the reverse would be slowing down SGTM's responsiveness in cases where the lambda does error/timeout. 🤔 I'll go with whatever you recommend.

suzyng83209 (approved):

terraform/sgtm-cluster/main.tf
vn6: hm I'll bump it down to 4 minutes so there's a little bit of extra time, and I wouldn't expect things to be sitting in queue for much longer

PR #174, opened 2024-04-29
Fix error "undefined check run completion time"

vn6:
### Summary

- check if completion time is undefined
- add unit test for this case

Asana tasks:
https://app.asana.com/0/1200740774079340/1206916572348053/f

Relevant deployment: 

CC: @vn6

### Risks

Pull Request: https://github.com/Asana/SGTM/pull/174

### DIFF
--- a/test/github/test_logic.py
+++ b/test/github/test_logic.py
@@ -746,6 +746,34 @@ class TestMaybeRerunStaleRequiredChecks(unittest.TestCase):
         )
         mock_merge_pull_request.assert_not_called()
 
+    @patch("src.github.logic.SGTM_FEATURE__CHECK_RERUN_ON_APPROVAL_ENABLED", True)
+    @patch("src.github.logic.SGTM_FEATURE__CHECK_RERUN_THRESHOLD_HOURS", 1)
+    @patch(
+        "src.github.logic.SGTM_FEATURE__CHECK_RERUN_BASE_REF_NAMES",
+        ["main"],
+    )
+    def test_no_rerun_when_check_run_not_complete(self, mock_rerequest_check_run):
+        check_run = build(builder.check_run().completed_at(None))
+        pull_request = build(
+            builder.pull_request()
+            .base_ref_name("main")
+            .commit(
+                builder.commit()
+                .status(Commit.BUILD_SUCCESSFUL)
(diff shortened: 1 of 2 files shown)


test/github/test_logic.py
```diff
@@ -746,6 +746,34 @@ def test_rerun_stale_checks_on_automerge_pr(
...
         mock_merge_pull_request.assert_not_called()
 
+    @patch("src.github.logic.SGTM_FEATURE__CHECK_RERUN_ON_APPROVAL_ENABLED", True)
+    @patch("src.github.logic.SGTM_FEATURE__CHECK_RERUN_THRESHOLD_HOURS", 1)
+    @patch(
+        "src.github.logic.SGTM_FEATURE__CHECK_RERUN_BASE_REF_NAMES",
+        ["main"],
+    )
+    def test_no_rerun_when_check_run_not_complete(self, mock_rerequest_check_run):
+        check_run = build(builder.check_run().completed_at(None))
```
suzyng83209, line 756: hmm surprisingly, there's no type error from mypy. The Check Run Builder's method for `completed_at` is `Union[str, datetime]` so you'll want to update that to be optional as well. Just for correctness

suzyng83209 (changes requested):

test/github/test_logic.py
vn6: updated the types!

test/impl/builders/check_run_builder.py
```diff
@@ -15,9 +15,11 @@ def __init__(
         }
 
     def completed_at(
-        self, completed_at: Union[str, datetime]
+        self, completed_at: Union[str, datetime, None]
     ) -> Union["CheckRunBuilder", CheckRun]:
-        self.raw_check_run["completedAt"] = transform_datetime(completed_at)
+        self.raw_check_run["completedAt"] = (
+            transform_datetime(completed_at) if completed_at else None
```
suzyng83209, line 21: You shouldn't need to do this ternary because `transform_datetime` handles the `None` case (although perhaps accidentally). 

Just an fyi. Feel free to leave this here or update the arg type of transform datetime

vn6: hm `transform_datetime` is used for other timestamps that we expect to be non-null so keeping this as is

PR #179, opened 2024-05-01
[SGTM performance] Modify comment webhook to be have fewer mutations

suzyng83209:
### Summary

- issue comments have 3 supported actions: created, edit, and deleted. We lock on the issue id (pull request id) and to a full `update_task` on the pull_request body when we don't need to
- instead of full updating, only upsert the comment and add followers if there are any
- instead of locking on the pr id, lock on the comment id

Asana tasks:
https://app.asana.com/0/1200740774079340/1207209762057512/f]
https://app.asana.com/0/1200740774079340/1207209762057512/f

Relevant deployment: 

CC: @prebeta @suzyng83209 @vn6 @michael-huang87 @harshita-gupta

### Risks

Pull Request: https://github.com/Asana/SGTM/pull/179

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1207215638657145)

### DIFF
--- a/src/github/controller.py
+++ b/src/github/controller.py
@@ -62,10 +62,7 @@ def upsert_review(pull_request: PullRequest, review: Review):
     pull_request_id = pull_request.id()
     task_id = dynamodb_client.get_asana_id_from_github_node_id(pull_request_id)
     if task_id is None:
-        logger.info(
-            f"Task not found for pull request {pull_request_id}. Running a full sync!"
-        )
-        # TODO: Full sync
+        logger.error(f"Task not found for pull request {pull_request_id}. Exiting!")
     else:
         logger.info(
             f"Found task id {task_id} for pull_request {pull_request_id}. Adding review"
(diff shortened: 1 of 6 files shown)


src/github/controller.py
```diff
@@ -53,15 +53,11 @@ def upsert_comment(pull_request: PullRequest, comment: Comment):
     pull_request_id = pull_request.id()
     task_id = dynamodb_client.get_asana_id_from_github_node_id(pull_request_id)
     if task_id is None:
-        logger.info(
-            f"Task not found for pull request {pull_request_id}. Running a full sync!"
+        logger.error(
+            f"Task not found for pull request {pull_request_id}. Exiting!"
         )
-        # TODO: Full sync
```
suzyng83209, line 59: doing a full sync is kind of difficult at this point in the code so instead of logging that we're doing something we actually aren't, I'm just logging an error. This doesn't seem to happen that often

prebeta: I'm a little worried about not syncing tasks. I feel like a complaint people have is not having data from a PR. Would exiting here mean there can be cases where a PR could exist without an associated task?

suzyng83209: There are definitely instances where a PR could exist without an associated task but full syncing was never properly implemented so there's no functional difference (i.e. this has always been an issue).

We do want some kind of full upsert on a task if the task id does not exist but that will have to come later

prebeta (approved):
LGTM!

suzyng83209:
test comment

PR #190, opened 2024-06-21
[Merge Queues] Fix automerge to work with merge queues

vn6:
### Summary

Turns out merge queues broke sgtm automerge since we needed to change the merge method :(

Asana tasks:

Relevant deployment: 

CC: @prebeta @suzyng83209

### Test Plan

note: I already deployed the changes in this PR to fix sgtm, tested with https://github.com/Asana/codez/pull/260823

### Risks

Pull Request: https://github.com/Asana/SGTM/pull/190

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1207634623287071)


src/github/client.py
```diff
@@ -42,7 +42,7 @@ def merge_pull_request(owner: str, repository: str, number: int, title: str, bod
     # we add the PR number to match Github's default squash and merge title style
     # which we rely on for code review tests.
     title_with_number = f"{title} (#{number})"
-    pr.merge(commit_title=title_with_number, commit_message=body, merge_method="squash")  # type: ignore
```
vn6, line 45: as per https://docs.github.com/en/graphql/reference/input-objects#enablepullrequestautomergeinput
image

merge_method is ignored when using a merge queue

suzyng83209: Does this still enable automerge to work in directories that don't have merge queues turned on? The docs aren't super clear 🤔

vn6: hmm looks like it will have the same functionality as clicking the "Click Enable auto-merge" button on the github UI https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request

I don't think we have this button turned on for codez but should work for repositories that do have that option

prebeta: We should verify that this works with repositories that have direct merge configuration as well.

vn6: currently testing in the testing-sgtm repo

prebeta (approved):
LGTM assuming testing goes well!

Thanks for looking into this!

src/github/client.py
vn6: looks like this won't work for repos without "automerge" turned on, updated to manual merge if trying to automerge fails (test run https://github.com/Asana/testing-sgtm/pull/4)

prebeta (approved):
LGTM

PR #192, opened 2024-06-24
[Merge Queues] Add merge queue status

vn6:
### Summary

- add "queued" status to indicate when PR is in merge queue, status will change back to "open" on eviction or "merged" after merge

Asana tasks:
https://app.asana.com/0/1203954674308315/1207270814999856/f

Relevant deployment: 

CC: @prebeta @suzyng83209

### Test Plan

- tested with https://app.asana.com/0/1203954674308315/1207650446738621/f (didn't get a screenshot in time before it changed to merged but it worked)

### Risks

Pull Request: https://github.com/Asana/SGTM/pull/192

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1207651182135382)

### DIFF
--- a/src/asana/helpers.py
+++ b/src/asana/helpers.py
@@ -85,6 +85,8 @@ def default_due_date_str(reference_datetime: Optional[datetime] = None) -> str:
 def _task_status_from_pull_request(pull_request: PullRequest) -> str:
     if pull_request.closed():
         return "Merged" if pull_request.merged() else "Closed"
+    elif pull_request.is_in_merge_queue():
+        return "Queued"
     else:
         return "Draft" if pull_request.is_draft() else "Open"
 
@@ -122,7 +124,7 @@ _custom_fields_to_extract_map = {
 def _custom_fields_from_pull_request(pull_request: PullRequest) -> Dict:
     """
     We currently expect the project to have three custom fields with its corresponding enum options:
-        • PR Status: "Open", "Draft", "Closed", "Merged"
+        • PR Status: "Open", "Draft", "Closed", "Queued", "Merged"
         • Build: "Success", "Failure"
         • Review Status: "Needs Review", "Changes Requested", "Approved", "Not Ready"
     """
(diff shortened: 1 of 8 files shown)


test/github/test_logic.py
```diff
@@ -442,6 +442,26 @@ def test_is_pull_request_ready_for_automerge_after_tests_mergeable_is_false(
...
+                .submitted_at("2020-01-13T14:59:58Z")
+                .state(ReviewState.APPROVED)
+            )
+            .isInMergeQueue(True)
+            .mergeable(MergeableState.CONFLICTING)
+            .merged(False)
+            .label(builder.label().name(github_logic.AutomergeLabel.AFTER_TESTS.value))
+        )
+        self.assertFalse(github_logic.maybe_automerge_pull_request(pull_request))
+        mock_merge_pull_request.assert_not_called()
```
suzyng83209, line 463: can you add another test case for the inverse of this test case? i.e. is in merge queue and not conflicting so mock_merge_pull_request is called?

suzyng83209 (commented):
Can you add another test plan for:
* repo with merge queue and
* project without the `Queued` CF

And are we interested in changing the build status if the pull request was rejected? Or are we fine with the failure slack message relaying that info?

src/asana/helpers.py
```diff
@@ -122,7 +124,7 @@ def _author_asana_user_id_from_pull_request(pull_request: PullRequest) -> Option
 def _custom_fields_from_pull_request(pull_request: PullRequest) -> Dict:
     """
     We currently expect the project to have three custom fields with its corresponding enum options:
-        • PR Status: "Open", "Draft", "Closed", "Merged"
+        • PR Status: "Open", "Draft", "Closed", "Queued", "Merged"
```
prebeta, line 127: Nice 👍

prebeta (approved):
LGTM

prebeta:
Does the "Queued" custom field need to be manually added to any project that uses SGTM?

vn6:
> And are we interested in changing the build status if the pull request was rejected? Or are we fine with the failure slack message relaying that info?

Not sure if we can easily get this information from graphql, might need to rely on the webhook metadata but I can do some additionally testing here. 

> Does the "Queued" custom field need to be manually added to any project that uses SGTM?

Yep, I updated the shared field that's being used in SGTM codez tasks and and 45 other projects which I think should cover everything we're using? 

image

PR #193, opened 2024-06-27
[MergeQueues] Delete branch if PR is merged

prebeta:
### Summary

Github's delete branch logic doesn't work if SGTM added the pull request to the merge queue. This will make SGTM delete branches if they haven't been cleaned up yet.

Asana tasks:
https://app.asana.com/0/1125126232217429/1207661000934486/f

Relevant deployment: 

CC: @suzyng83209 @prebeta @vn6 @michael-huang87

### Test Plan

Run on test sgtm deployment

### Risks

Pull Request: https://github.com/Asana/SGTM/pull/193

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1207680708784404)


src/github/client.py
```diff
@@ -60,3 +60,28 @@ def rerequest_check_run(owner: str, repository: str, check_run_id: int):
...
+    """
+    try:
+        repo = gh_client.get_repo(f"{owner}/{repo_name}")
+        # Attempt to get the branch, will raise a 404 error if not found
+        repo.get_branch(branch_name)
+        ref = f"heads/{branch_name}"
+        git_ref = repo.get_git_ref(ref)
+        git_ref.delete()
+        logger.info(f"Branch '{branch_name}' deleted successfully.")
+    except Exception as e:
```
skeggse, line 82: i would imagine this is a `GithubException` where `from github.GithubException import GithubException`

src/github/client.py
```diff
@@ -1,7 +1,7 @@
 import requests
 from requests.auth import HTTPBasicAuth
 
-from github import PullRequest  # type: ignore
+from github import PullRequest, GithubException  # type: ignore
```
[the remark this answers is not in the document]
prebeta: Ooh interesting. Let me try this and remove the type ignores

src/github/logic.py
```diff
@@ -296,6 +296,14 @@ def maybe_automerge_pull_request(pull_request: PullRequest) -> bool:
     return False
 
 
+def maybe_delete_branch_if_merged(pull_request: PullRequest):
+    if pull_request.merged():
```
suzyng83209, line 300: Just to note: even though this is called after `maybe_automerge`, the `pull_request` state at this point will not be merged if it's handling the same webhook event. This is because `maybe_automerge` doesn't (currently) update the state of the pull request object. Instead, it will fire on the next webhook event that gets triggered on the `close` action. Just an fyi while testing.

suzyng83209 (changes requested):

src/github/models/pull_request.py
```diff
@@ -219,3 +219,6 @@ def labels(self) -> List[Label]:
 
     def base_ref_name(self) -> str:
         return self._raw["baseRefName"]
+
+    def head_ref_name(self) -> str:
+        return self._raw["headRefName"]
```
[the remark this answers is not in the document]
prebeta: yeah i just added it into src/github/graphql/fragments/FullPullRequest.py

src/github/logic.py
prebeta: yup, i think it makes sense that this will only happen on the PR close action.

src/github/webhook.py
```diff
@@ -18,6 +18,8 @@ def _handle_pull_request_webhook(payload: dict) -> HttpResponse:
         pull_request = graphql_client.get_pull_request(pull_request_id)
         # a label change will trigger this webhook, so it may trigger automerge
         github_logic.maybe_automerge_pull_request(pull_request)
+        if payload["action"] == "closed":
```
prebeta, line 21: Put this check in so we don't delete branches on arbitrary PR updates.

suzyng83209: I suppose someone could create a branch with a new name, then re-open the same pull request and merge it which would result in a deletion but this seems like an unlikely situation so 👍

suzyng83209 (changes requested):
code looks good. Can you add new test cases in `test/github/test_webhook.py` and `test/github/test_logic.py`? You may need to create a new field in `test/impl/builders/pull_request_builder.py` which is the mock pull request builder object to supply pull request objects to tests

suzyng83209 (approved):
nice lgtm and thanks for fixing types

PR #194, opened 2024-07-01
Add branch deletion exceptions

michael-huang87:
### Summary

https://github.com/Asana/SGTM/pull/193 started deleting "main" branches when PRs were being made from that branch. Add some logic to ignore deletion logic for certain branches. We could maybe make this configurable through an env variable but not sure how useful that would be

Asana tasks:
https://app.asana.com/0/1202309366996843/1207707608547616/f

### Test Plan

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1207708659445679)

### DIFF
--- a/src/github/graphql/client.py
+++ b/src/github/graphql/client.py
@@ -7,6 +7,7 @@ from .queries import (
     GetPullRequestByRepositoryAndNumber,
     GetPullRequestAndComment,
     GetPullRequestAndReview,
+    GetRepository,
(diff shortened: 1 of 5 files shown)


suzyng83209 (changes requested):
I'm not a huge fan of this change because its hardcoding branch names that may or may not be protected. You can deploy this as a temporary fix to unblock IT or deploy a version of SGTM without #193 but I'm going to talk to IT to see if we can find a better solution

PR #196, opened 2024-07-17
Remove code AOR banner

dikarel:
Code AORs aren't active in this repo, so it can be confusing to readers

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1207835561826737)


suzyng83209 (approved):
Thanks for the deleting. I think this was actually added due to running `git pr` on a non CODEZ repo. Is there source code in `git pr` that could be modified to account for public repos not getting the CodeAor banner?

PR #208, opened 2025-04-22
feat: Add support for native GitHub media attachments 

savokr-asana:
**Description**
This PR introduces support for adding github attachments to corresponding asana task. Currently, it only happens on PR creation, will be fixed for PR updates as well soon.

**Test**
Added unit tests are passing together with `extract_attachments` function tested on real codez PR and video/images attachments.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -43,69 +43,84 @@
         },
         "certifi": {
             "hashes": [
-                "sha256:3cd43f1c6fa7dedc5899d69d3ad0398fd018ad1a17fba83ddaf78aa46c747516",
-                "sha256:ddc6c8ce995e6987e7faf5e3f1b02b302836a0e5d98ece18392cb1a36c72ad56"
+                "sha256:3d5da6925056f6f18f119200434a4780a94263f10d1c21d032a6f6b2baa20651",
+                "sha256:ca78db4565a652026a4db2bcdf68f2fb589ea80d0be70e03929ed730746b84fe"
             ],
             "markers": "python_version >= '3.6'",
-            "version": "==2024.6.2"
+            "version": "==2025.1.31"
         },
         "cffi": {
             "hashes": [
-                "sha256:0c9ef6ff37e974b73c25eecc13952c55bceed9112be2d9d938ded8e856138bcc",
-                "sha256:131fd094d1065b19540c3d72594260f118b231090295d8c34e19a7bbcf2e860a",
-                "sha256:1b8ebc27c014c59692bb2664c7d13ce7a6e9a629be20e54e7271fa696ff2b417",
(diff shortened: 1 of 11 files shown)


savokr-asana:
Just found out one thing that attachments copy-pasted from local files are mention in HTML  tag even in markdown, so will need to add support for that as well, it will be quite easy and I will try to simplify current logic to rely only on html based text body instead of markdown.

suzyng83209 (approved):
Sorry for missing this pull request for a while. All of the code looks good. I do think you need to fix some types but the test failures look transient so a re-run may resolve it.

Let me know when you want to deploy this change!

savokr-asana:
@suzyng83209 Thanks for the review! There is one thing still missing in the PR, and it's update of the attachments when the  github description/comment is updated. It's really important since the flow with `git pr` forces us to add attachments only after the initial PR is created, so in order to actually add them properly in asana, we should support the update of the task as well. I will try to work on it tomorrow and you can probably re-review the changes and try testing it as well.

src/aws/lock.py
```diff
@@ -1,16 +1,40 @@
...
+    return _dynamodb_lock_client
+
+
+# Create a lazy-loading module-level variable
+class _LazyLockClient:
+    def __getattr__(self, name):
+        return getattr(get_dynamodb_lock_client(), name)
+
+
+dynamodb_lock_client = _LazyLockClient()
```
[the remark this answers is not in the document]
vn6: Makes sense and looks good to keep - though I don't think it'll be huge improvement since we always need to lock client to be fully set up whenever we access it.

test/impl/mock_dynamodb_test_case.py
```diff
@@ -4,14 +4,17 @@
...
-import boto3  # type: ignore
-from moto import mock_dynamodb  # type: ignore
-from src.config import AWS_REGION, OBJECTS_TABLE, LOCK_TABLE
-from .base_test_case_class import BaseClass
 
 # "Mock" the AWS credentials as they can't be mocked in Botocore currently
 os.environ.setdefault("AWS_ACCESS_KEY_ID", "foobar_key")
 os.environ.setdefault("AWS_SECRET_ACCESS_KEY", "foobar_secret")
+# Set ENV to test so the config module provides proper test defaults
+os.environ.setdefault("ENV", "test")
```
vn6, line 12: are we using this env var anywhere?

vn6 (approved):

test/impl/mock_dynamodb_test_case.py
savokr-asana: Good catch, no, forgot to remove these changes after figuring out how to fix CI errors. This file shouldn't have any changes at all, fixed that.

PR #213, opened 2025-07-02
adding more options for s3 terraform backend

vignirhafsteinsson:
The [Terraform S3 backend ](https://developer.hashicorp.com/terraform/language/backend/s3) supports state locking by default without using DynamoDB for locks. In fact, [that option is deprecated and will be removed in a future version](https://developer.hashicorp.com/terraform/language/backend/s3#state-locking). However, I did not want to change the default behaviour of the state backend so it is still the preferred option.

Also added the option of managing the key prefix from an environment variable.

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1210697536520263)
Pull Request: https://github.com/Asana/SGTM/pull/213

### DIFF
--- a/terraform/terragrunt.hcl
+++ b/terraform/terragrunt.hcl
@@ -16,10 +16,15 @@ terraform {
 %{ else }
   backend "s3" {
     bucket = "${get_env("TF_VAR_terraform_backend_s3_bucket_name")}"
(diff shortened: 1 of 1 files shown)


vn6 (approved):
lgtm but we're actually primarily using the terraform cloud backend instead of the s3

PR #214, opened 2025-08-18
[Github attachments sync] Do not include unwanted attachments for PR comments

savokr-asana:
This PR implements "blacklisting" certain sources for attachments to remove unwanted images in the asana task.

Asana task:
https://app.asana.com/1/15793206719/project/1208527156432726/task/1210630232406671?focus=true

CC:
@suzyng83209 @vn6 

Pull Request synchronized with [Asana task](https://app.asana.com/0/0/1211082770903846)
Pull Request: https://github.com/Asana/SGTM/pull/214

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -67,6 +67,7 @@ module "sgtm-prod" {
   sgtm_feature__autocomplete_enabled             = var.sgtm_feature__autocomplete_enabled
   sgtm_feature__automerge_enabled                = var.sgtm_feature__automerge_enabled
   sgtm_feature__automerge_disabled_repositories  = var.sgtm_feature__automerge_disabled_repositories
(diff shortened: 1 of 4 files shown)


suzyng83209 (changes requested):
When we deploy, we actually populate the environment variables with terraform so we need to make modifications to `terraform/main.tf`, `terraform/variables.tf`, `terraform/sgtm-cluster/lambda_function.tf`, and `terraform/sgtm-cluster/vars.tf`.

and FYI: variables.tf is where we define that we expect terraform to prompt for these values from the user for the entire workspace. vars.tf does it for the lambda_function module. Lambda_function specifies how those variables will be passed to the underlying terraform resource blocks. main.tf contains a list of lambda_function modules that we define for the workspace (prod and staging). The `SGTM_FEATURE__AUTOMERGE_DISABLED_REPOSITORIES` flag should have a similar set up already!

suzyng83209 (approved):

PR #224, opened 2026-04-16
Fix empty string in excluded attachment URLs config

harshita-gupta:
## Context

SGTM syncs GitHub PR images and videos into Asana tasks as attachments. In PR #214 (Aug 2025, @savokr-asana), a feature was added to exclude attachments from configurable URL sources (e.g., to filter out unwanted tracker pixels or badge images). This is controlled by the `SGTM_FEATURE__EXCLUDED_ATTACHMENT_SOURCES_URLS` env var — a comma-separated list of hostnames to block.

That feature introduced a config parsing bug: when the env var is **unset** (the default for local dev and any deployment that doesn't configure it), `"".split(",")` produces `[""]`, so the excluded-sources set contains a single empty string `""`. Since `str.endswith("")` and `"" in str` are always `True` in Python, **every attachment URL matches the exclusion list**, and `_extract_attachments()` silently returns `[]` — no attachments are ever synced.

## How we found it

`test/asana/helpers/test_extract_attachments.py` has 7 pre-existing test failures on master. Running `pipenv run python -m pytest test/asana/helpers/test_extract_attachments.py -v` shows all 7 return empty lists where attachments are expected. These failures were introduced by PR #214 but went unnoticed because:
1. `pytest` isn't in the Pipfile, so devs may not run these tests locally
2. There's no visible CI test runner in the repo (the only check is Graphite mergeability)
3. The `ENV=test` config path (which overrides the buggy default) may mask the issue in some environments

## What's affected

- **Local dev**: all attachment extraction is broken when `SGTM_FEATURE__EXCLUDED_ATTACHMENT_SOURCES_URLS` is unset
- **Production**: any deployment using the Terraform default (`""`) for this variable has the same bug — attachments are silently dropped. Deployments that set it to a real value (e.g., `"tracker.example.com"`) are unaffected.
- The bug is in `src/config.py` line 39, not in the extraction logic itself (`src/asana/helpers.py`), which is correct.

## Fix

Filter out empty strings when parsing the env var, using a set comprehension with `if url`. This matches the pattern already used for `SGTM_FEATURE__FOLLOWUP_REVIEW_GITHUB_USERS` at line 53 of `src/config.py`.

Also add a `@patch` to `test_extract_attachments_with_excluded_source_url` so it explicitly mocks the config value instead of depending on the `ENV=test` override.

## Note for reviewers

`SGTM_FEATURE__AUTOMERGE_DISABLED_REPOSITORIES` (line 36) has the same `set("".split(","))` → `{""}` pattern, but it's **not exploitable** because its consumer uses `repo_name in set` (equality), not substring/endswith matching. I left it alone to keep this PR minimal.

## Test plan
- [x] All 10 tests in `test_extract_attachments.py` pass (previously 7 failed)
- [x] Full test suite (214 tests) passes with no regressions

🤖 Generated with [Claude Code](https://claude.com/claude-code)


vn6 (approved):
lgtm

